PR #3, opened 2019-03-23
Remove deprecated JVM option UseConcMarkSweepGC

ocisly:
Hi there, I came across your project and took it for a spin 🙌

This patch fixes the following warning (printed after exiting neovim):

    OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.

We shouldn't need to specify a particular garbage collector implementation; let the JVM choose.

(You probably don't need `CMSClassUnloadingEnabled` either – it's ignored by the newer garbage collectors – but it doesn't do any harm.)

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -19,4 +19,4 @@
   {:extra-paths ["classes"]
    :jvm-opts ["-client" "-Xmx2g" "-Xverify:none"
               "-XX:+TieredCompilation" "-XX:TieredStopAtLevel=1"
-              "-XX:+UseConcMarkSweepGC" "-XX:+CMSClassUnloadingEnabled"]}}}
+              "-XX:+CMSClassUnloadingEnabled"]}}}


Olical:
Ah wonderful, thanks a lot! I'm no expert when it comes to optimising the JVM for startup time, these options were taking from a wiki page on benchmarks for the boot build tool. I think it was pretty old so it makes sense that they're outdated.

Weird that I didn't see the warning though, I guess it's my JVM version. I'll drop the other one too, thanks for the tips!

Olical:
Feel free to use the latest commit for now, it should work nicely, I'm just getting inner and outer form eval selection working with `[]` and `{}` pairs before the next tagged version. I fixed the `make compile` running every time even when it hasn't actually updated too.

PR #26, opened 2019-04-17
Stop JVM gracefully

jlesquembre:
Every time I quit neovim, I get a message on the terminal: `Conjure exited, restarting`, and the process never terminates, I need to kill it with `CTRL-C`, but it leaves a zombie java process on my system.

Debugging it, I found that conjure always exits with code 143. I think it means that the application was terminated due to a SIGTERM command.

I'm using openjdk version "1.8.0_202".

I took the `clean-up-and-exit` function from [nrepl](https://github.com/nrepl/nrepl):
https://github.com/nrepl/nrepl/blob/2b432c859c5a98b6e8c97338315bfaab2961641b/src/clojure/nrepl/cmdline.clj#L17-L23

### DIFF
--- a/src/conjure/main.clj
+++ b/src/conjure/main.clj
@@ -11,9 +11,18 @@
             [conjure.ui :as ui]
             [conjure.action :as action]))
 
+(defn- clean-up-and-exit
+  "Performs any necessary clean up and calls `(System/exit status)`."
+  []
+  (shutdown-agents)
+  (flush)
+  (binding [*out* *err*] (flush))
+  (System/exit 0))
+
 (defn -main
   "Start up any background services and then wait forever."
   []
+  (.addShutdownHook (Runtime/getRuntime) (Thread. #(clean-up-and-exit)))
   (dev/init)
   (rpc/init))
 


Olical:
Ah this is neat, thank you very much! I'm just really confused as to why you see this and I don't 🤔 a serious case of "works on my machine".

Maybe the OS or something... happy to merge this though! Thanks!

jlesquembre:
I use linux, not sure if that is relevant. But it's strange that you don't see the error if you are also using openjdk 1.8.
Anyways, thanks for merging it, the error message was annoying me

Olical:
Oh I'm on Java 11 OpenJDK! Oh Arch Linux, so maybe it's just different JDK
versions. I think someone else reported it too at some point. Thank you for
fixing it!

Olical:
Hmm, doesn't look like the JVM is shutting down for me despite `System/exit` being called. How odd :thinking: I have to `killall -s 9 java` between `bin/dev` executions. I guess something is preventing it from stopping but I'm not sure what.

jlesquembre:
Strange, and did it work before my change?
Didn't try to use `bin/dev` but for me the JVM was stopped after closing neovim

Olical:
Yep, it was fine before, how odd. I can't work it out right now, I guess one of the futures of servers is holding it open but I'm not sure why `System/exit` isn't stopping them. I thought that stopped everything.

Olical:
Looks like it hangs in the normal compiled run too, not just dev :thinking:

```
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-b01)
OpenJDK 64-Bit Server VM (build 25.212-b01, mixed mode)
```

So why does mine not like this and yours doesn't like the lack of it I wonder.

Olical:
https://github.com/Olical/conjure/commit/e980791a0cd51531df552b120fa880f5c351d20a I've been a little firmer with the "please shutdown now!" and it seems okay for me. I think something was deadlocking? This is probably better, I'd rather Conjure always exited completely no matter what.

Issue #7, opened 2019-03-26
Configuration and mappings


Olical:
1. Work out how best to configure things.
 2. Configure things.

Things like horizontal splitting and only showing the last result as mentioned in #4 would be easy to add. I guess like `let g:conjure_default_mappings = 0`, just define global variables and Conjure will look for them.

What else could be configured easily enough :thinking:

PR #35, opened 2019-05-15
Make Conjure log window size and direction configurable

jlesquembre:
Related to #7, I wanted to see how looks like if Conjure's window log opens in an horizontal split, and wrote this code.

It adds 3 new variables, `g:conjure_log_direction`, `g:conjure_log_size_small` and `g:conjure_log_size_large` to define the direction and size of the Conjure log window.

Also modifies the size meaning. Before the size was the number of columns to use, now it is the percentage of the window to use. 
To me, makes more sense to make it a percentage, since sometimes I work on a small monitor (laptop) and other times on a bigger one.

### DIFF
--- a/lua/conjure.lua
+++ b/lua/conjure.lua
@@ -23,8 +23,26 @@ local function find_log (log_buf_name)
   return nil
 end
 
+
+-- Returns the absolute number of lines
+local function get_log_window_size(size, direction)
+  local percentage_size = (size == "small") and
+    vim.api.nvim_get_var("conjure_log_size_small") or
+    vim.api.nvim_get_var("conjure_log_size_large")
+
+  local vim_window_size = (direction == "horizontal") and
+    vim.api.nvim_get_option("lines") or
+    vim.api.nvim_get_option("columns")
+
+  return math.floor(vim_window_size * (percentage_size / 100))
+end
+
+
 -- Find or create (and then find again) the log window and buffer.
-function conjure.upsert_log (log_buf_name, width, focus, resize)
+function conjure.upsert_log(log_buf_name, size, focus, resize)
+  local direction = vim.api.nvim_get_var("conjure_log_direction")
+  local size_abs = get_log_window_size(size, direction)
+
   local result = find_log(log_buf_name)
   if result then
     if focus == true then
@@ -32,12 +50,17 @@ function conjure.upsert_log (log_buf_name, width, focus, resize)
     end
 
     if resize == true then
-      vim.api.nvim_win_set_width(result.win, width)
+      if direction == "horizontal" then
+        vim.api.nvim_win_set_height(result.win, size_abs)
+      else
+        vim.api.nvim_win_set_width(result.win, size_abs)
+      end
     end
 
     return result
   else
-    vim.api.nvim_command("botright " .. width .. "vsplit " .. log_buf_name)
+    local split = (direction == "horizontal") and "split" or "vsplit"
+    vim.api.nvim_command("botright " .. size_abs .. split .. " " .. log_buf_name)
     vim.api.nvim_command("setlocal winfixwidth")
     vim.api.nvim_command("setlocal buftype=nofile")
     vim.api.nvim_command("setlocal bufhidden=hide")
--- a/src/conjure/nvim/api.clj
+++ b/src/conjure/nvim/api.clj
@@ -64,8 +64,22 @@
   {:method :nvim-buf-set-var
    :params [buf (util/kw->snake name) value]})
 
+(defn get-var
+  [name]
+  {:method :nvim-get-var
+   :params [(util/kw->snake name)]})
+
 (defn set-var [name value]
   {:method :nvim-set-var
+   :params [(util/kw->snake name) value]})
+
+(defn get-option
+  [name]
+  {:method :nvim-get-option
+   :params [(util/kw->snake name)]})
+
+(defn set-option [name value]
+  {:method :nvim-set-option
    :params [(util/kw->snake name) value]})
 
 (defn execute-lua [code & args]
--- a/README.md
+++ b/README.md
@@ -91,6 +91,12 @@ You can disable these and define your own with `let g:conjure_default_mappings =
 
 If you get something wrong it'll explain using [Expound][] in the log buffer. Essentially you must provide at least a `:tag` and `:port`.
 
+### Options
+
+ * `g:conjure_log_direction` - How to split the log window. `"vertical"` or `"horizontal"`. Defaults to `"vertical"`.
+ * `g:conjure_log_size_small` - Percentage size of the window used by Conjure log. A number between 1 and 100, defaults to 25.
+ * `g:conjure_log_size_large` - Percentage size of the window used by Conjure log **when open with `ConjureOpenLog`**. A number between 1 and 100, defaults to 50.
+
 ## Completion
 
 Completion is provided through the wonderful [Compliment][] and is injected for you without conflicting with existing versions. `<c-x><c-o>` omnicompletion should work as soon as you're connected.
--- a/autoload/conjure.vim
+++ b/autoload/conjure.vim
@@ -8,6 +8,10 @@ else
   let s:job_opts = "-A:fast"
 endif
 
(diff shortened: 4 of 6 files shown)


Olical:
Oh this is fantastic! I was going to address your points in #7 eventually but you beat me to it :tada: I'll give it a review now, I've found myself wanting a horizontal split regularly too.

I'm not 100% (hah) sure about percentage sizes but I'll give it a think. Maybe a choice makes sense between a number and a percentage. I like knowing my split will be exactly wide enough to fit all of the formatted output inside it.

I suppose 25% and 50% would be more than enough on a 1080p monitor though :thinking: I was thinking of bumping the large size to 100 columns today, coincidentally.

lua/conjure.lua
```diff
@@ -24,20 +24,28 @@ local function find_log (log_buf_name)
...
+function conjure.upsert_log (log_buf_name, size, direction, focus, resize)
   local result = find_log(log_buf_name)
   if result then
     if focus == true then
       vim.api.nvim_set_current_win(result.win)
     end
 
     if resize == true then
-      vim.api.nvim_win_set_width(result.win, width)
+      if direction == "horizontal" then
```
Olical: I'd recommend `==#` here even though it doesn't really matter. If we want to be case insensitive we should use `==?`. Using `==` is okay but will listen to the users `ignorecase` setting inside their Neovim. It shouldn't be a problem here but it's better to remove Vim's ambiguity where we can.

src/conjure/nvim/api.clj
```diff
@@ -64,10 +64,24 @@
...
    :params [(util/kw->snake name) value]})
 
+(defn get-option
+  [name]
+  {:method :nvim-get-option
+   :params [(util/kw->snake name)]})
+
+(defn set-option [name value]
+  {:method :nvim-set-option
+   :params [(util/kw->snake name) value]})
```
Olical: Good job working out how I was structuring the API and higher level `conjure.nvim` side! :clap:

You're the first person to touch this code other than me so let me know if anything was surprising or bad, happy to tidy things up if we both agree there's a better way :smile:

lua/conjure.lua
```diff
@@ -24,20 +24,28 @@ local function find_log (log_buf_name)
...
     end
 
     return result
   else
-    vim.api.nvim_command("botright " .. width .. "vsplit " .. log_buf_name)
+    if direction == "horizontal" then
+      vim.api.nvim_command("botright " .. size .. "split " .. log_buf_name)
+    else
+      vim.api.nvim_command("botright " .. size .. "vsplit " .. log_buf_name)
+    end
```
Olical: The comment about `==#` or `==?` over `==` applies here too. Just keeps it working consistently across different configurations.

How about we simply choose the split command with a ternary operator like this:

```viml
let l:split = direction ==# "horizontal" ? "split" : "vsplit"
vim.api.nvim_command("botright " .. size .. split .. " " .. log_buf_name)
```

Olical:
To follow on from that slightly, if we defined the initial config in `plugin/conjure.vim` like this:

```viml
let g:conjure_log_direction = get(g:, 'conjure_log_direction', "vertical")
let g:conjure_log_size_small = get(g:, 'conjure_log_size_small', 25)
let g:conjure_log_size_large = get(g:, 'conjure_log_size_large', 50)
```

It would allow us to default the values while also letting users override the config before or after Conjure has been loaded. It's a neat trick I think :slightly_smiling_face:

jlesquembre:
Thanks for your comments. I understand better now how things are working.
Keep in mind that I have no Lua experience, just let me know if the code is unidiomatic or looks bad, feedback is always welcome :)

About the percentage sizes, maybe we can make the size options accept a string, like `"50%"`. E.g.: if the value is `50` (a number) or `"50"` (a string) the size means number of lines/columns, while if the value is `"50%"` it means a percentage. This way we can choose between a number and a percentage without introducing a new option. What do you think?

About moving the functionality from Clojure to Lua, it makes a lot of sense, I didn't think about performance (and I also prefer to write Clojure ;) ) . But something is not completely clear to me, I understand that calling Neovim from Clojure blocks the process. But what happens when calling Neovim from Lua (e.g.: `vim.api.nvim_win_set_width`)? Is it blocking too?

You mention that you have wider thoughts about configuration. I can address your feedback and update my PR, but if you have other ideas about how to implement it, it's totally fine if you close this PR and you do it in a different way. Just let me know how to prefer to proceed.

lua/conjure.lua
jlesquembre: I'm new to Lua, but is `==#` valid Lua? Maybe you thought that it was viml code while is Lua?

src/conjure/nvim/api.clj
jlesquembre: I found that code relatively easy to understand, it looks good to me, I don't have any improvement suggestion

lua/conjure.lua
jlesquembre: As mention before, maybe you are confusing Lua and Viml? Not completely sure about it, but looks like in Lua there is no ternary operator (or at least I cannot find anything on google)

lua/conjure.lua
Olical: I THOUGHT THIS WAS VIML.

I'm an idiot :laughing:

lua/conjure.lua
Olical: Yep, again, thought it was VimL, great observation skills there. Maybe we could at least set a local variable (with lua I think `foo = bar` is global and `local foo = bar` is local to your block?) with the split method then use that. Just didn't want that whole `nvim_command` call repeated really.

Olical:
I'm learning lua as I go along with this too, so that's fine. I prefer to write Clojure too of course, that's why the "find matching pairs" code is in Clojure because it's quite complex. So I minimised the amount of calls it made (two I think?) then built all of the logic in Clojure, I've had to be super careful with the performance of it though, batching and reusing as much as possible.

I may well translate that to lua some day if benchmarks (when I write them) prove that `read-form` is one of the slowest parts of Conjure, which it may well be! I don't look forward to trying to write that in lua though. Worth noting that I think the lua in Neovim runs a lot faster than the VimL, so I would favour lua over VimL for perf.

Using the lua does prevent porting to regular vim though which might be a "fun" modification to make later...

So I think calling Neovim from lua ends up being a direct call in memory, luajit is built into the neovim binary which is cool. There's no RPC / msgpack encode and decode which is what eats up time when Conjure's JVM and Neovim communicate. There appears to be almost no cost to that.

I think the way you designed the config looks good to me and how I was going to approach it, I'd just set defaults using the `get` method I mentioned in one of my replies so users can override before or after they're set but we can rely on there always being a value in those `g:` variables.

With regards to percentages, I think just stick with numbers for now and it'll always be a percent. Then _maybe_ we add a toggle at some point that disables the percentage part. I think it's a better way to do it so we'll swap to percentages for now.

Sorry about writing so much in response to these, just want to provide context and hopefully help you decide more things / understand why I'm thinking one way or another :slightly_smiling_face:

jlesquembre:
It definitely helps me understand, I appreciate long responses with more context, thanks :)

I will update my PR. FYI, I'm not sure if I'll have time for it this week, if not I'll do it next week

Olical:
No problem, take your time. Calm development is good development. I'll be working on Conjure most of this weekend I think trying to get sustainable dependency injection working for Compliment and future dependencies like Orchard.

If you want me to take a look over the weekend I'll be happy to merge your changes and tweak them as we discussed. Happy to leave it until you're done though so you can claim the lines :smile:

jlesquembre:
For me, if you want to work on Conjure during the weekend, and you'd like to finish this feature, go for it. 
Don't merge my changes if they are not needed, fine with me, no claims ;)
On the other hand, if you prefer to focus on other features, I'm happy to work on this PR.

jlesquembre:
I just updated the PR. I think I addressed all your feedback, but let me know if I forgot something.

There is no ternary operator in Lua, but I found that the same can be accomplish with `$condition and ... or ...`

Olical:
All merged, let's test this on master for a bit. I think maybe I'm going to need a develop branch soon...

Issue #10, opened 2019-03-26
Changed namespace reloading


Olical:
This depends on your project including [tools.namespace](https://github.com/clojure/tools.namespace) (a pretty common dependency) Again, there's some old code from the legacy branch that can be used to learn from.

https://github.com/Olical/conjure/blob/ff0f4a18e5b37a5cd44d6d0bae28653e3c90e3ea/autoload/conjure.vim#L108-L131

Olical:
Currently on develop, releasing in the next version!

PR #36, opened 2019-05-19
Mr Anderson usage

Olical:
Well this was a fun one to write! This PR replaces the manual fetching and require order calculation for compliment with Mr Anderson and tools.namespace.

What does this mean for the user? Hopefully, nothing (yet). You shouldn't see any differences since it's doing the same thing, I did replace `make compile` with `bin/compile` though so people will have to be aware of that in the next release.

Side note: I upgraded all the CircleCI usage to run two jobs for the compile and tests in parallel.

This work will let me add any injected dependencies I want easily, so now I can add namespace reloading bindings (#10)! I can ensure the right version of tools.namespace is always loaded as you connect.

I'll also be able to add orchard and cljs-tools in the near future so I can delete some of my hacky go to definition code.


bbatsov:
> I'll also be able to add orchard and cljs-tools in the near future so I can delete some of my hacky go to definition code.

Happy to hear this! In Orchard 0.5 (currently in beta) we've merged the var-info ClojureScript features  from cljs-tooling, but cljs-tooling remains a good options for ClojureScript code completion. We're also working on pushing this aspect of cljs-tooling into Compliment and deprecating cljs-tooling (see https://github.com/alexander-yakushev/compliment/pull/62). Generally the plan is for Orchard to provide feature parity for Clojure and ClojureScript functionality. Feedback and suggestions for improvements would be most welcome. I really want us to develop an API that's useful to all dev tool maintainers.

Olical:
Yep, that compliment issue is also on my radar! My plan is to replace a bunch of my code with orchard (cljs-tooling where required) and hopefully compliment on the Clojure and ClojureScript side. Conjure will slowly turn into the glue / platform to plug these things together which will be nice!

I'm following the progress closely, hopefully once I'm a consumer I can join in with contributions, thanks for all of the great work in this space :+1: feature parity for Orchard would be :drooling_face:

PR #37, opened 2019-05-21
Remap and config

### LINKED ISSUE #7: Configuration and mappings

Olical:
Closes #7, does the following:

 * Moves code out of `autoload` into `plugin`, there was no point in having this in two files anyway.
 * Add support for `.edn` files so you can eval those data structures. You can do it by replacing your `:expr` in your connection but this does it by default.
 * Updated the mappings to use some more of the keyboard to make room for more mappings soon, looking at the diff is the best way to see what I've done there. Of note will be things like `,re`/`,rr` moving to `,ee`/`,er` for eval expression and eval root.
 * Added some config values with defaults for the changes @jlesquembre is making as well as for disabling the default mappings. Hopefully that's something you were going to do in your PR anyway! :slightly_smiling_face:

Feedback appreciated!


README.md
```diff
@@ -45,16 +44,16 @@ The `'for'` and `'on'` keys are optional but you might prefer Conjure to only st
...
+ * `<localleader>ee` - `ConjureEvalCurrentForm`
+ * `<localleader>er` - `ConjureEvalRootForm`
+ * `<localleader>ee` - `ConjureEvalSelection` (visual mode)
+ * `<localleader>eb` - `ConjureEvalBuffer`
+ * `<localleader>ef` - `ConjureLoadFile`
+ * `<localleader>cs` - `ConjureStatus`
+ * `<localleader>cl` - `ConjureOpenLog`
+ * `<localleader>cq` - `ConjureCloseLog`
+ * `<localleader>tt` - `ConjureRunTests`
+ * `<localleader>ta` - `ConjureRunAllTests`
```
Olical: * `e` is anything eval related.
 * `t` is anything test related.
 * `c` is anything meta to do with Conjure itself.

I'll probably add `f` for formatting things and `r` for reloading things with `tools.namespace` soon.

Olical: Maybe `cc` for "clojure log" and `co` or `cl` for open log :thinking:

Issue #32, opened 2019-05-07
Show current function signature 


Olical:
Not sure how or when, maybe on `CursorHold` and `CursorHoldI` it should echo the signature. Showing the signature in completion results would be nice too, I'll have to see if that's possible.

The main point being: The function signature for what you're currently working with should just be there, the doc lookup binding should be for when you need even more information. I find myself calling doc repeatedly just to get the argument order, I need to make that process smoother.

Olical:
Will be released in v0.15.0.

PR #38, opened 2019-05-28
Quick doc

Olical:
This addresses #32, so now when `CursorHold` fires you'll see a snippet of the documentation for the current form's head symbol in the echo messages area.

You can control the delay with `set updatetime=500` for a 500ms delay before it looks anything up. I need to use it for a while to make sure it doesn't cause weird errors preventing you from typing, that'll be super annoying.

Feel free to give this branch a whirl!

## To Do

 * [x] Work out why some macro forms don't get doc - **reason**: `read-string` doesn't like autoresolved maps and keywords like `::q/foo` or `#::q/foo{...}`. We can't read those strings without having all of the aliases defined from the target namespace - **fix**: Use `tools.reader` with some extra options so we can parse more code.
 * [x] Reproduce and fix the root cause of the rare error where `split-lines` is called on a `Cons` instead of a `String`. (have only seen this once or twice, may have fixed it by mistake...) - **reason**: Looks like it's because the doc lookup call is coming in while the prelude is still evaluating which is a bit weird - **fix**: I guess block these evaluations until everything is done?

## Maybe Eventually

 * Get it working in insert mode, `CursorHoldI` seems kinda broken in Neovim https://github.com/neovim/neovim/issues/3757 (might have to implement my own debounce)
 * Standardise for doc and quick doc that it tries to check the symbol under the cursor _and then_ the head of the current form.
 * Allow doc of keywords if they contain specs. (should really be var, special form or spec keyword)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You can find out more about socket prepls in my blog post, [Clojure socket prepl
  * Running tests in the current namespace or the `-test` equivalent.
  * Friendly error output by default with optional expansion.
  * Go to definition.
- * `(doc ...)` lookup on `K` or via `CursorHold` for your current form.
+ * Documentation lookup on a key press and when your cursor is idle.
 
 ### Upcoming
 
@@ -44,7 +44,7 @@ The `'for'` and `'on'` keys are optional but you might prefer Conjure to only st
 You can disable these and define your own with `let g:conjure_default_mappings = 0`.
 
  * `InsertEnter` in a Clojure buffer (that is _not_ the log) closes the log.
- * `CursorHold` in a Clojure buffer looks up the docs for the head of the form under your cursor and echos it. If this is annoying you, turn it off with `let g:conjure_quick_doc = 0`
+ * `CursorHold` in a Clojure buffer looks up the docs for the head of the form under your cursor and displays it with virtual text.
  * `<localleader>ee` - `ConjureEvalCurrentForm`
  * `<localleader>er` - `ConjureEvalRootForm`
  * `<localleader>ee` - `ConjureEvalSelection` (visual mode)
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Here's an example with [vim-plug][], my plugin manager of choice.
 Plug 'Olical/conjure', { 'tag': 'v0.14.2', 'do': 'bin/compile', 'for': 'clojure', 'on': 'ConjureAdd'  }
 ```
 
-You should rely on a tag so that breaking changes don't end up disrupting your workflow, please don't depend on `master`. Make sure you watch the repository for releases using the menu in the top right, that way you can upgrade when it's convenient for you.
+You should rely on a tag so that breaking changes don't end up disrupting your workflow, please don't depend on `master` (and especially not `develop`!). Make sure you watch the repository for releases using the menu in the top right, that way you can upgrade when it's convenient for you.
 
 The `'for'` and `'on'` keys are optional but you might prefer Conjure to only start up once you've entered a Clojure file.
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You can find out more about socket prepls in my blog post, [Clojure socket prepl
  * Running tests in the current namespace or the `-test` equivalent.
  * Friendly error output by default with optional expansion.
  * Go to definition.
- * `(doc ...)` lookup.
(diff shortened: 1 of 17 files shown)


Olical:
:eyes:

Screenshot_2019-05-29_13-14-00

I'll be doing the same with one line eval outputs that have no stdin/out too (you can still open the log buffer if you want!). This way your regular `def` evals etc won't constantly open the log buffer which can be kinda costly since it requires a lot of back a forth.

I don't want virtual text to be the _only_ place you see information because I don't like the fact that I can't interact with it. But if the information fits I want it there as a convenience thing to avoid having to look around my screen too much.

@jlesquembre if you have any thoughts on this, feel free to chime in! I value your opinion, if you're still using Conjure, that is :sweat_smile:

Olical:
I seem to be able to get it into a weird state sometimes but a restart or a reconnect fixes it. It seems to try and split the lines of a cons, which I presume means something is returning an error, not a string. I would _guess_ it's to do with `read-form` pulling in more than one form or some unbalanced parenthesis.

Next time I reproduce it I'll put the stack here, I can pave over the issue easily but I'd like to fix the root cause.

jlesquembre:
I'm definitely still using Conjure, in fact, I replaced fireplace with Conjure. 
While you wrote the last comment, I was just trying this branch :) 
I was thinking the same, I find this feature really useful, but I didn't like having to look to bottom of my screen all the time, I really like your idea of using virtual text.

I can't think of a better solution, good job!

If I see any errors or a weird behavior, I'll post it here

Olical:
That's extremely motivating to hear! Thank you! But also no hard feelings if you ever drop back to anything else, I just appreciate your feedback while you continue to use it :slightly_smiling_face:

I know there's a bug here somewhere, but when I reconnect or restart it seems to fix itself, so I think it's maybe a REPL getting into a weird state with an open paren or something (yep, that can happen with a prepl...). Once I've reproduced and fixed that I'll be happy to get this out.

Then I'll try to do something kinda similar for small eval output to avoid _always_ opening the log buffer for tiny outputs. I guess that'll replace the quick doc output when you eval... I don't think you'll want both at once and I'll like to keep it to one virtual text at a time on screen.

Olical:
So I fixed a bug over lunch around parsing code containing unknown namespace aliases for autoresolving maps like `#::x{:y :z}`, it'll now resolve that `x` alias to `user`. This shouldn't be a problem in practice since `parse-code` is used internally by Conjure for stuff like doc lookup.

And I somehow reproduced that weird error!

```
Error from thread 'RPC message handler':
 {:cause "clojure.lang.PersistentList cannot be cast to java.lang.CharSequence",
 :trace [[clojure.string$split invokeStatic "string.clj" 219]
         [clojure.string$split_lines invokeStatic "string.clj" 228]
         [conjure.action$quick_doc invokeStatic "action.clj" 86]
         [conjure.main$fn__15434 invokeStatic "main.clj" 89]
         [conjure.main$fn__15434 invoke "main.clj" 89]
         [clojure.lang.MultiFn invoke "MultiFn.java" 229]
         [conjure.rpc$init$fn__14905$fn__14908 invoke "rpc.clj" 198]
         [clojure.core$binding_conveyor_fn$fn__5739 invoke "core.clj" 2030]
         [clojure.lang.AFn call "AFn.java" 18]
         [java.util.concurrent.FutureTask run "FutureTask.java" 266]
         [java.util.concurrent.ThreadPoolExecutor runWorker
          "ThreadPoolExecutor.java" 1149]
         [java.util.concurrent.ThreadPoolExecutor$Worker run
          "ThreadPoolExecutor.java" 624]
         [java.lang.Thread run "Thread.java" 748]],
 :via
   [{:at [clojure.string$split invokeStatic "string.clj" 219],
     :message
       "clojure.lang.PersistentList cannot be cast to java.lang.CharSequence",
     :type java.lang.ClassCastException}]}
```

 1. Started Neovim
 2. Opened a file with cursor on a form
 3. Connected the REPL
 4. Saw doc output
 5. Moved around
 6. Now getting that error on most (**but not all**) doc lookups

Reconnecting to the REPL fixes it, so I think something is out of sync and there's some error result that is mixing up the `:ret` values I'm getting out of the prepl. Maybe the doc lookup is firing before the prepl is ready.

Olical:
Alright, happy with this now! Fixed the race condition bug by removing the buffers on the eval, read and ret channels, they weren't required and allowed things to get out of sync. Might just pop the buffers back in for ret/read but keep it limited for eval.

PR #42, opened 2019-06-11
Optionally fold multiline results

justone:
I really like the way that errors are folded, and I deal with many large results, so I thought allowing folding of multiline results would be useful. Turns out it was quite easy to add.

It's fun to work on a Vim plugin written in Clojure.


Olical:
Looks good! Probably not something I'll use (at the moment?) but I can see others wanting it. I'm going to rename `g:conjure_fold_results` to `g:conjure_fold_multiline_results` though, just so it's a little more specific and is in line with the wording used by `auto_open`. I hope that's okay!

justone:
The name change sounds great. Thanks for merging!

PR #45, opened 2019-06-19
Feat/add plug bindings

dharrigan:


README.md
```diff
@@ -111,6 +111,14 @@ If you open Conjure without any connections configured it'll self prepl into it'
  * `K` - `ConjureDoc`
  * `gd` - `ConjureDefinition`
 
+### Remapping
+
+It's possible to remap the keys, for example, to remap the `ConjureToggleLog`, add the following to your `.vimrc`:
+```
+ nmap <silent> <Leader>cl <Plug>(ConjureToggleLog)
+ nmap <silent> <Leader>co <Plug>(ConjureOpenLog) "remap the ConjureOpenLog to something else
```
Olical: Hmm so the thing I don't like about this is that they're not buffer local whereas all of the default mappings are. The way I was going to get around this was with a key mapping data structure that gets merged with a bunch of defaults.

Not sure if this is the correct approach :thinking:

dharrigan: Hi,

That's just my example. In the code, there're buffer local. Let me update the readme.

dharrigan: Done

Olical: Oh, does that still work as an override? I would've thought then when Conjure's VimL kicks in it'll override anything from your `.vimrc`.

dharrigan: It seems to work for me. Give it a whirl.

Olical: Hmm, that example is now for the current buffer, not for Clojure buffers. My point being if you create a mapping under a Clojure autocmd then the mapping will only exist _after_ the buffer has been opened. The Conjure mappings are set up before that happens though unless you set Conjure to only boot in Clojure files, which you may well be doing.

I hope I'm explaining myself but I don't think I am :grimacing:

README.md
```diff
@@ -111,6 +111,14 @@ If you open Conjure without any connections configured it'll self prepl into it'
  * `K` - `ConjureDoc`
  * `gd` - `ConjureDefinition`
 
+### Remapping
+
+It's possible to remap the keys, for example, to remap the `ConjureToggleLog`, add the following to your `.vimrc`:
+```
+ nmap <silent> <buffer> <Leader>cl <Plug>(ConjureToggleLog)
+ nmap <silent> <buffer> <Leader>co <Plug>(ConjureOpenLog) "remap the ConjureOpenLog to something else
```
Olical: So these mappings are made for the first buffer you open and no more, if you prefix them with a Clojure autocmd then I don't think they'll override Conjure. Conjure will set up it's autocmds regardless because your mapping won't exist yet.

plugin/conjure.vim
```diff
@@ -53,30 +53,40 @@ command! -nargs=0 ConjureToggleLog call conjure#notify("toggle_log")
 command! -nargs=* ConjureRunTests call conjure#notify("run_tests", <q-args>)
 command! -nargs=? ConjureRunAllTests call conjure#notify("run_all_tests", <q-args>)
 
+function conjure#mapKey(plugMapping, keyMapping, action)
+    if !hasmapto(a:plugMapping)
+        exec 'autocmd FileType clojure nnoremap <silent> <buffer>' a:keyMapping a:action
+    endif
+    exec 'autocmd FileType clojure nnoremap <silent> <buffer>' a:plugMapping a:action
```
Olical: Is this second exec a typo? :thinking:

plugin/conjure.vim
```diff
@@ -53,30 +53,40 @@ command! -nargs=0 ConjureToggleLog call conjure#notify("toggle_log")
...
+    :call conjure#mapKey("<Plug>(ConjureLoadFile)", "<localleader>ef", ":ConjureLoadFile <c-r>=expand('%:p')<cr><cr>")
+    :call conjure#mapKey("<Plug>(ConjureUp)", "<localleader>cu", ":ConjureUp<cr>")
+    :call conjure#mapKey("<Plug>(ConjureStatus)", "<localleader>cs", ":ConjureStatus<cr>")
+    :call conjure#mapKey("<Plug>(ConjureOpenLog)", "<localleader>cl", ":ConjureOpenLog<cr>")
+    :call conjure#mapKey("<Plug>(ConjureCloseLog)", "<localleader>cq", ":ConjureCloseLog<cr>")
+    :call conjure#mapKey("<Plug>(ConjureToggleLog)", "<localleader>cL", ":ConjureToggleLog<cr>")
+    :call conjure#mapKey("<Plug>(ConjureRunTests)", "<localleader>tt", ":ConjureRunTests<cr>")
+    :call conjure#mapKey("<Plug>(ConjureRunAllTests)", "<localleader>ta", ":ConjureRunAllTests<cr>")
+    :call conjure#mapKey("<Plug>(ConjureDoc)", "K", ":ConjureDoc <c-r><c-w><cr>")
+    :call conjure#mapKey("<Plug>(ConjureDefinition)", "gd", ":ConjureDefinition <c-r><c-w><cr>")
```
Olical: Although I've raised other concerns, won't be needing that `:` prefix on `call` I don't think. Also since every mapping ends in `<cr>` might as well make that part of the `mapKey` function.

Olical:
I am still pretty convinced this'll have to be data driven by merging a (potentially undefined) user map (as in the data structure) with some defaults. Then building the mapping autocmds from that data. It's less flexible from a Vim Script perspective but will allow us to give proper configuration.

dharrigan:
Hi,

I'm not convinced by having the keybindings in the edn configuration (if that is what you mean). All the plugins I use (37!!) do not store their key bindings externally in a configuration, so it's sort of breaking the convention of having the bindings in the vim script with user overrides in each individual's vimrc. Why deviate from that? It's pretty normal practice to override settings in the vimrc (I'm sure even you do that :-))

Of course, as the owner of the plugin, you are 100% free to do what you think is best for the future, however, it won't be "standard" if there is a thing ;-)

Olical:
Oh I'm not suggesting that at all, my bad if it came across that way. I
mean storing it in a vim script map that gets merged with defaults.

Olical:
I'm going to close this for now because it requires a bunch of VimL (I think?) to configure every mapping, but I'll use the ideas from it! Will try to get this done soon. Thanks for the thoughts and ideas!

PR #47, opened 2019-07-04
Result refactor

### LINKED ISSUE #40: Records can't be serialised and deserialised by Conjure
`mynomoto` over on Slack pointed out that if you try to return a record from an eval Conjure explodes.

```clojure
(defrecord Foo [bar])
(->Foo "bar")
```

Because that evaluates to the string

```clojure
#conjure.code.Foo{:bar \"bar\"}
```

Which Conjure tries to read because it reads all prepl responses (for better or worse...). I think the main need to read responses is because I want to check if it was an error or not. That will go away when I overhaul how errors and exceptions work.

Would be cool to fix this ASAP, but if you hit this before then just `(into {} r)` where `r` is your record. It'll get around this issue until I can patch it out or it vanishes on it's own when I improve errors.

Olical:
Mostly done! This branch greatly improves how errors are caught and handled within Conjure. I've made sure caught exceptions get logged too which should make debugging in the future easier when someone breaks Conjure. The whole plugin _should_ be more robust in weird edge cases now, time for me to test it out for a while!

A few more small things:

 * [x] Remove use of `conjure.result` from `conjure.code/parse-ns`
 * [x] Delete all traces of `conjure.result`
 * [x] Ensure that this fixes #40, I'm 99% confident that it does


Olical:
Yeah, whatever, codecov. You can whine all you want but I know it's okay. I'll get you a high test coverage percentage some day, my angry GitHub integration :heart:

PR #61, opened 2019-09-14
Add ability to show source for a symbol

justone:


plugin/conjure.vim
```diff
@@ -92,6 +93,7 @@ augroup conjure
     call s:def_config("nmap_refresh_changed", g:conjure_map_prefix . "rr")
     call s:def_config("nmap_refresh_all", g:conjure_map_prefix . "rR")
     call s:def_config("nmap_refresh_clear", g:conjure_map_prefix . "rc")
+    call s:def_config("nmap_source", g:conjure_map_prefix . "ss")
```
Olical: I'll give the mapping some hammock time 🤔 I don't mind it at all, seems as good as any, but I'll let it sit for a bit before committing to it 😄 

This is a really cool addition and done really well, thank you very much! I'm glad you managed to find all the little nuances of what you have to do to add new commands! And worked out the templating system!

src/conjure/ui.clj
```diff
@@ -89,6 +89,11 @@
   [{:keys [conn resp]}]
   (append {:origin (:tag conn), :kind :doc, :msg (:val resp)}))
 
+(defn source
+  "Results from a (source ...) call."
+  [{:keys [conn resp]}]
+  (append {:origin (:tag conn), :kind :source, :msg (:val resp)}))
```
Olical: Does this render commented out? I haven't tried it yet, would be cool if it would print like a multi line ret with syntax highlighting etc.

justone: Yes it does. Is there a way to return it as source and have it highlighted?

plugin/conjure.vim
justone: Yeah, I went with ss for "show source", but if you think of a different one, let me know.

Yeah, the implementation is a bit copy/paste heavy from the doc command, but I didn't want to start over generalizing too quickly.

Olical: I think a little copy / paste is okay when we're not sure if we'll over DRY ourselves, it's manageable.

Olical: I like it, I might add `<localleader>sd` for show docs too, as well as `K`.

Olical:
Thank you so much for the contribution! I'm just tweaking it to render it as code and get it working for unqualified symbols. I realised it won't work on something like `foo` but will on `bar/foo`, I'll get it sorted and show you what I changed.

Olical:
btw, to render as code you can just pass `:code? true` to `append` :D

PR #62, opened 2019-09-18
Improve the wording of the message that prints when conjure exits unexpectedly

daveyarwood:
This error message confused me initially. Now that I have context, I figured there was an opportunity to improve the error message a bit. Feedback welcome!


Olical:
Nice! I like it :D thank you very much, it's definitely an improvement!

PR #80, opened 2020-04-20
Update documentation

mamapitufo:
* Use the correct git revision for the installation instructions
* Update help entry tag to match the `README`


Olical:
Good spot! Thanks! I thought it had to be the filename but that makes a lot more sense, I copied that from another help doc that was wrong too I guess. Cheers! :tada:

PR #85, opened 2020-05-13
Fix typo in README.adoc

svanburen:
Conjure is excellent, keep it up 😄


Olical:
Thanks for the fix and the kind words!

PR #96, opened 2020-05-25
Main help file improvements

mamapitufo:
* Fix incorrect "it's".
* Remove repetition of "configuration".

Feel free to drop the stylistic fix; it's just that when I re-read the paragraph, I noticed the multiple "configuration" and thought it could be a quick improvement.


Olical:
Hey! Thank you very much! I like the changes, it definitely reads better. I still mess up "it's" and "its" in certain scenarios :grimacing: you'd think I would get it eventually.

PR #99, opened 2020-06-10
Add timer feature for HUD (hide HUD some time after cursor move)

theHamsta:
I found it quite stressful that the HUD would disappear immediately after CursorMove.

This adds a time-out for the HUD to close only 1 second after CursorMove. An additional eval during this period defuses an already running timer.

If this feature is desired I can add an option for the length of the time-out and only activate it when `defer_fn` is available (nightly).

### DIFF
--- a/fnl/conjure/log.fnl
+++ b/fnl/conjure/log.fnl
@@ -8,6 +8,8 @@
             text conjure.text
             editor conjure.editor}})
 
+(var current-timer-id 0)
+
 (defonce- state
   {:hud {:id nil}})
 
@@ -28,6 +30,14 @@
         (nvim.win_close state.hud.id true)
         (set state.hud.id nil)))))
 
+(defn defer-close-hud []
+  (let [my-id current-timer-id]
+    (when state.hud.id
+      (vim.defer_fn (fn []
+                      (when (= my-id current-timer-id)
+                        (close-hud)))
+                    1000))))
+
 (defn- break-lines [buf]
   (let [break-str (break)]
     (->> (nvim.buf_get_lines buf 0 -1 false)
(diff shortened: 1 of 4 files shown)


dharrigan:
Hi,

I like the HUD disappearing immediately on cursor movement. I don't mind this option, but make it opt-in please :-)

-=david=-

Olical:
> I just wanted to show a demo to test if this feature would ever an option.

Ah, okay! Understood.

> make this configurable, default off and only on available when defer_fn is available

Yep, definitely this. If you want, I can merge your code as is and handle that part for you? I could also look into finding a solution that doesn't rely on `defer_fn`? That way your contribution and commits are in the repo but I can tweak the surrounding code for you? Saves you the effort of adding config and docs which I have bunch of experience with. Up to you!

I think this PR is a great start, I just want some config and docs added as well as a dig into "can this be done in stable Neovim easily?" because I feel like I've done it before :thinking: are you happy for me to merge it and tweak it into a release, or would you like to address these things some time soon in your own time?

theHamsta:
Hi. I guess I left most of this PR open (how to configure (vim global, fennel variable)? how to do handle backwards compatibility?) because I didn't know about your personal preferences for this project.

Since the feature is rather small it would probably be a lot more effective if you would finish it.

Thank you a lot for including it in the project! It's really a game changer for me!

Olical:
Fair enough, totally understand! I'll get this merged into develop and do the padding around it to add config etc, you'll be able to see what I did in those commits for some more context too. Not much to it, but it's still something that is mostly just in my head and the code right now.

Olical:
Okay, merged, configured and tweaked to work with Neovim stable on `develop`. Use the following line in your vimrc (or at runtime) to adjust the delay from the default of `0` (off).

```vim
ConjureConfig log.hud.passive-close-delay 1000
```

Olical:
Hmm a running timer keeps Neovim from fully exiting :thinking: calling stop doesn't seem to free it, investigating.

Olical:
Okay, I'm destroying the timer properly now, should be all good. Let me know what you think! I'll close this PR but feel free to tack more on if you have issues or it works great and you think it can be released :D :tada:

PR #106, opened 2020-07-24
Change clojure test to use test-vars

b-ryan:
This switches the Clojure test runner command so it uses `clojure.test/test-vars` instead of `clojure.test/test-var`. Using `test-vars` causes any fixtures to be used in the test run.

If it helps, this is what vim-fireplace does. Here's a PR for that I found https://github.com/tpope/vim-fireplace/pull/231

I don't exactly know how to test this, but the change seemed straightforward enough :man_shrugging:


Olical:
Merged to `develop`! Looks good to me! Thank you :smile:

PR #107, opened 2020-07-24
Make enhanced cljs completion optional

hpointu:
It seems a recent version of clj-suitable broke cljs completion,

see https://github.com/rksm/clj-suitable/issues/15 for more details

Having clj-suitable optional (from config) is an ok workaround, until this is fixed.

This PR introduces a conf : `g:conjure#client#clojure#nrepl#completion#cljs#use_suitable`, to control the use of clj-suitable.

This conf value is also initialized to `false` by default, because as it stands, putting it to true wouldn't work.


Olical:
I think when this is fixed and okay in the future I'll flip the default to `true`, but `false` is the right call for now.

PR #122, opened 2020-10-22
Add docs for evaluating Clojure code in markdown

seantwie03:
@Olical you mentioned this setting that allows evaluation of Clojure code from inside of a markdown file on my Reddit thread yesterday.  I thought it might be useful to add it to the docs.  You already cover this filetype_client stuff in [conjure.txt](https://github.com/Olical/conjure/blob/0e63269e176232cdd9f1c20d94dfd3b8ad5702a1/doc/conjure.txt#L196) but having it here might be a bit easier for people to find.   Feel free to  modify this or close it without merging if you disagree.

### DIFF
--- a/doc/conjure-client-clojure-nrepl.txt
+++ b/doc/conjure-client-clojure-nrepl.txt
@@ -308,4 +308,9 @@ All configuration can be set as described in |conjure-configuration|.
             machines in large buffers with lots of top level forms.
             Default: `false`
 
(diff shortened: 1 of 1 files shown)


Olical:
Thanks! Yeah I think it'll be useful, not sure if this will be the final resting place of this info but we'll see. I'd maybe move it to some heading that stands alone, or put it in the wiki :thinking: if you ever write a post about this sort of thing (evaluating within other kinds of content) let me know and I'll link to it from the wiki :smile:

PR #155, opened 2020-12-24
WIP: add keymap/function to fetch/present usage examples.

tami5:
This PR attempts to introduce new functionality that enables users to
preview usage examples from https://clojuredocs.org (and perhaps down the line
https://janetdocs.com and other similar sites for conjure clients).

**How it works:**

It downloads a gist/remote file to `~/.cache/conjure`, then parses the
content to fnl map/dict. Then, through helper function it further parse
the dict into a markdown string and append to a temp markdown buffer
(e.g. `conjure-usage-686895.md`). Finally, It offer two options/mappings
to view the buffer, either via a popup or split.

**Selling point**

View community usage examples/notes and enhance your understanding
of high-level function from within youe editor.

**Limitation & Drawbacks**

This is list of few things I consider as drawbacks to the
implementation:

- It downloads a file to user filesystem.
- It downloads it from a gist.
- It isn't fully asynchronous, there slight lag/block when parsing json/fnl (help needed).

**Things to improve on**

- Modularity and the ability to easily clients support.
- Callbacks.
- Skipping reading/parsing client defs files.
- Messages.

**Todos**

- [X] get clojuredocs.org data
- [x] format markdown buffer.
- [x] get a symbol namespace under the cursor.
- [x] create split/vsplit and float display function.
- [ ] create mappings and commands.
- [ ] refactor: use filetype to determine client 
- [ ] refactor: use conjure.promise  (awaiting @Olical review)
- [ ] update docs (awaiting @Olical confirmation on whether to include this feature in conjure)

**Won't do in this PR**

- Enable navigation between symbol docs through `See Also` section.
- Add support for janetdocs.org

**Testing**

```clj
;; Markdown parsing and formating result.
((. (require :conjure.usage) :get-sym-usage)
  :clojure :clojure.core/map
  #(let [path "/tmp/clojure.map.md"]
      (a.spit path $1)
      (vim.cmd (.. "edit " path))))

(get-usage
    {:client :clojure
     :display (or :float :vsplit :split)
     :symbol "map"
     :fill 0.7
     :win {:winblend 0}})
```

```lua
lua require'conjure.usage'['get-sym-usage']("clojure", "clojure.core/map", function(content) require'aniseed.core'.spit("/tmp/clojure.map.md", content); vim.cmd(.. "edit /tmp/clojure.map.md") end)
```

### DIFF
--- a/fnl/conjure/usage.fnl
+++ b/fnl/conjure/usage.fnl
@@ -1,0 +1,241 @@
+(module conjure.usage
+  {require {nvim conjure.aniseed.nvim
+            fennel conjure.aniseed.fennel
+            a conjure.aniseed.core
+            str conjure.aniseed.string
+            config conjure.config
+            conjure-client conjure.client
+            eval conjure.eval
+            extract conjure.extract
+            fs conjure.fs}})
+
+(def clojure
+  {:url
+   (.. "https://gist.githubusercontent.com" "/tami5/"
+       "14c0098691ce57b1c380c9c91dbdd322" "/raw/"
+       "b859bd867115960bc72a49903e2b8de0ce249c31" "/"
+       "clojure.docs.fnl")
+   :path (.. (fs.cache-dir) "/" "clj-docs.fnl")
+   :ft "fennel"})
+
+(defn- get-msg [client mtype]
+  "Error msg pased on error type."
+  (let [f string.format]
+    (match mtype
+      :not-supported (f "'%s' client is not supported" client)
+      :fetch-err (f "CONJURE-ERROR: %s usage doc could not downloaded.
+                  Try again, or open an issue." client)
+      :fetching (f "CONJURE: fetching usage defs for %s ......." client)
+      :caching (f "CONJURE: caching usage defs for %s ......." client))))
+
+(defn- get-in-client [client key]
+  "Retrun a client key value."
+  (. (match client
+       :clojure clojure
+       (error (get-msg client :not-supported))) key))
+
+(defn- dl-defs [client cb]
+  "Download doc usage file."
+  (let [path (get-in-client client :path)]
+    (if (fs.exists? path) (cb)
+      (let [args ["curl" "-L" (get-in-client client :url) "-o" path]
+            on_exit #(if (fs.exists? path) (cb)
+                       (error (get-msg client :fetch-err)))]
+        (print (get-msg client :fetching))
+        (vim.fn.jobstart args {:on_exit on_exit})))))
+
+(defn- up-defs [client cb]
+  "Redownload client usage defs/file."
+  (vim.fn.jobstart
+    ["rm" (get-in-client client :path)]
+    {:on_exit #(dl-defs :clojure cb)}))
+
+(defn- cached-defs? [client]
+  "Returns true if client.cache is nil."
+  (not (a.nil? (. (. *module* client) :defs))))
+
+(defn- cache-defs [client cb]
+  "Fill client runtime cache with usage defs."
+  (let [path (get-in-client client :path)
+        parse (match (get-in-client client :ft)
+                "fennel" #(fennel.dofile path)
+                "json" #(vim.fn.json_decode (vim.fn.readfile path)))]
+    (dl-defs client
+              (vim.schedule_wrap
+                #(do (print (get-msg client :caching))
+                   (tset (. *module* client) :defs (parse))
+                   (cb))))))
+
+(defn- ensure-defs [client cb]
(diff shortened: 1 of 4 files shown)


fnl/conjure/usage.fnl
```diff
@@ -0,0 +1,133 @@
...
+  "Error msg pased on error type."
+  (let [f string.format]
+    (match mtype
+      :not-supported (f "'%s' client is not supported" client)
+      :fetch-err (f "CONJURE-ERROR: %s usage doc could not downloaded.
+                  Try again, or open an issue." client)
+      :fetching (f "CONJURE: fetching usage defs for %s ......." client)
+      :caching (f "CONJURE: caching usage defs for %s ......." client))))
+
+(defn- get-in-client [client key]
```
tami5, line 28: Does it make sense to skip the use of clients? and instead use filetype names to determine the client. I think this might be possible if client table {url, path, ft ...} will be take the client name.

tami5: But again I guess in its place will have vim.bo.filetype all over the place. It may make sense to abbreviate it to to a local variable?

fnl/conjure/usage.fnl
```diff
@@ -0,0 +1,133 @@
...
+       "b859bd867115960bc72a49903e2b8de0ce249c31" "/"
+       "clojure.docs.fnl")
+   :path (.. (fs.cache-dir) "/" "clj-docs.fnl")
+   :ft "fennel"})
+
+(defn- get-msg [client mtype]
+  "Error msg pased on error type."
+  (let [f string.format]
+    (match mtype
+      :not-supported (f "'%s' client is not supported" client)
```
tami5, line 22: @Olical please tell me if this looks good to you or suggest edits

fnl/conjure/usage.fnl
```diff
@@ -0,0 +1,241 @@
...
+  (print " ") ;; clear prompt
+  (vim.cmd "new")
+  (setup-buffer 0 opts.content opts.buf))
+
+(defn- open-vsplit [opts]
+  (print " ") ;; clear prompt
+  (vim.cmd "vnew")
+  (setup-buffer 0 opts.content opts.buf))
+
+(defn get-usage [opts]
```
tami5, line 208: I guess here we might want the use to know that we're trying to open usage for the symbol under the cursor : 

`Fetching community documentation for "symbol" ....`

fnl/conjure/usage.fnl
Olical: This shouldn't really be implemented in this central place, it should be part of the individual client modules such as `conjure.client.clojure.nrepl`.

This means that a 3rd party client or any new client in general can support this feature by exposing functions in their module's public interface, see `:help conjure-clients` for information on what's already in the expected client public interface already.

You should think of client modules as potentially not even in this repository, I want Conjure users to be able to combine 3rd party projects that slot into Conjure seamlessly, client modules are a core part of that idea.

fnl/conjure/usage.fnl
Olical: Not sure I fully understand the question, but no, I don't want to support two ways of dividing functionality and identifying what code should be used to handle actions, that's what the client modules and interface is designed for.

Olical:
If you don't think you can fit the implementation into the client modules (i.e. that design is just not right for this feature) I think it really reinforces the idea that this should be an optional plugin on the side. The fact that it fetches things from outside and writes to disk (something I've deliberately avoided) worries me slightly, mostly only because I have to look after this feature which isn't something I would choose to add personally.

I can help set up the repo (which could just be a fork of Conjure with a bunch deleted I guess) if you want? That way you have full ownership and ability to design and develop it however you want. I can then put it in the install steps for those that are interested (it'd be two lines to paste into your plugin list instead of one which is pretty easy).

The other benefit to making this a 3rd party repository that slots into Conjure easily that you own would be you have full freedom to iterate on it and develop it on your own terms. You can do whatever you see fit, make your ideal UX and design without my ideals getting in the way. I think I am going to end up restricting your vision because I like to keep everything within client modules and the log buffer UX.

fnl/conjure/usage.fnl
tami5: @Olical hmmm not sure I get what you mean or what you want me todo, anyway I'm moving the code to its own repo as plugin as you've suggested. Please next time be more clear on what actionable things/steps I need to do or take to make this better. Thanks for take the time to read to review

tami5:
https://github.com/tami5/conjure-clojuredocs

Issue #166, opened 2021-01-24
The fennel stdio client breaks Neovim when using some readline config


nicopap:
First: thank you for making conjure.

Using the fennel stdio client, I found that neovim would get into a weird state that can only finish in a crash.

Good news, with a default setup, this doesn't happen.

### How to reproduce

* install lua, fennel, lua readline via luarocks. Setup environment so that lua finds the luarocks module
* add `let g:conjure#filetype#fennel = "conjure.client.fennel.stdio"` to your neovim config
* write the following snippet in your `~/.inputrc`:

```
$if fennel
    set editing-mode vi
    set show-mode-in-prompt on
    set vi-ins-mode-string "\1\e[3 q\2\1\015\2fennel"
    set vi-cmd-mode-string "\1\e[1 q\2\1\015\2fennel"
$endif
```

* make sure fennel found the readline lib by opening the repl with `fennel` and seeing there is no messages about readline anymore
* open a fennel file with neovim
* ahah, neovim bork, only way out is Ctrl-C

### My workaround

The error comes from my exotic readline config. Since I only setup readline with lua5.1 and not other versions, I can fix this by using a different lua version for fennel by adding the `--lua /path/to/luaVer` command line switch to the fennel command. What I did:

`let g:conjure#client#fennel#stdio#command = "fennel --lua /usr/bin/lua5.4"`

It's questionable whether you need the default fennel repl when you have conjure. But it's useful to report this bug so that when someone stumble upon this, they have a chance of finding the solution in the conjure bug tracker instead of having to waste an hour figuring things out :P

### Solution

I can't think of a "good" solution to this. I have many workarounds that come to mind, but they all involve unexpected side effects that probably leads to more problems than solutions:
* Remove the readline library from the `LUA_PATH` when invoking fennel from neovim
* Do not install/uninstall the readline lua library
* I have my suspicions that what is breaking neovim are the terminal escape sequences, so filtering them out in conjure might help. But that's not a good solution

The **real solution** to this is to add to the fennel executable a `--no-funny-stuff` cli switch to disable any user-specific funny stuff and keep a consistent interface so that conjure and other software can rely on it. But that involves making an upstream change to fennel, and I'm very lazy. On top of that, fennel is probably too young to commit to a "consistant CLI for automatic consumption".

Well, not a very actionable bug report, I'm sorry. But I just wanted to make sure people with the same issues are not stuck with no hopes of fixing it.

Again, my thanks and accolades for making conjure.

[...]

PR #167, opened 2021-01-24
Document bug in stdio fennel client with readline

nicopap:
There is no easy solution to this until we can tell the fennel repl to
give us a client-friendly interface. So we document it in the hope that
the user can use the info to implement their own workaround. See #166


Olical:
Yeah, I think you're right, this is a good mitigation for now, but if the stdio stuff behaves weird there's not much I can do. Unless there's environment variables I can set to disable readline things before something is loaded?

Olical:
Thanks for the docs until we think of something better!

PR #169, opened 2021-02-01
Add config for `current_form_names`

mracos:
Which is used when checking if the current root form is a test. Before was a hardcoded "deftest" string :)
I see that this could be really specific for my context but maybe someone else might find it useful 😄

Obligatory: tks for the great plugin! 💙


fnl/conjure/client/clojure/nrepl/init.fnl
```diff
@@ -37,6 +37,9 @@
        :before nil
        :dirs nil}
 
+      :test
+      {:current_form_names ["deftest"]}
```
mracos, line 41: I'm not that happy with this config name but couldn't think of something that was better 🤷

Olical: Hmm I mean it makes sense in context, out of context less so but that's okay and expected. It's a kinda nuanced concept, so adding config is always a balancing act, I've had this feeling with most options I've added like "it doesn't 100% fit, but it's close".

I think it's good, but if I think of anything else I'll update it or let you know, I suspect I'll just go with this though :smile:

Olical:
This is done really well :smile: you got all the details right, nice one! Merged to `develop` for now, will cut a release with a few other changes after it's had time to soak. I use `develop` day to day at work and I think some more intrepid users rely on it too, so it gives us time to catch any issues.

Thanks a lot! Very tidy!

fnl/conjure/client/clojure/nrepl/init.fnl
mracos: That makes a lot of sense, great then!

mracos:
Thank you for the detailed next steps as well as for the kind compliments! 💙

PR #173, opened 2021-02-13
Prevent loading readline config on fennel stdio

### LINKED ISSUE #166: The fennel stdio client breaks Neovim when using some readline config

nicopap:
This fixes #166. The bug was caused by readline inserting control
characters into the fennel repl. By skipping the user readline config,
they are prevented from modifying the repl in arbitrary dangerous ways.
This is fine, because the readline editing facilities are completely
useless in the context of conjure, since we rely solely on neovim's
editing capabilities.

Note: The user is still able to configure the lua readline library
through their ~/.fennelrc file, for example, to persist their history.

### DIFF
--- a/doc/conjure-client-fennel-stdio.txt
+++ b/doc/conjure-client-fennel-stdio.txt
@@ -6,7 +6,6 @@ CONTENTS                                *conjure-client-fennel-stdio-contents*
     1. Introduction ........ |conjure-client-fennel-stdio-introduction|
     2. Mappings ................ |conjure-client-fennel-stdio-mappings|
     3. Configuration ...... |conjure-client-fennel-stdio-configuration|
-    4. Bugs ........................ |conjure-client-fennel-stdio-bugs|
 
 ==============================================================================
 INTRODUCTION                        *conjure-client-fennel-stdio-introduction*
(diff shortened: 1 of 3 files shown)


nicopap:
needs more testing

PR #174, opened 2021-02-13
Prevent loading readline config on fennel stdio

### LINKED ISSUE #166: The fennel stdio client breaks Neovim when using some readline config

nicopap:
This fixes #166. The bug was caused by readline inserting control
characters into the fennel repl. By skipping the user readline config,
they are prevented from modifying the repl in arbitrary dangerous ways.
This is fine, because the readline editing facilities are completely
useless in the context of conjure, since we rely solely on neovim's
editing capabilities.

Note: The user is still able to configure the lua readline library
through their ~/.fennelrc file, for example, to persist their history.

### DIFF
--- a/fnl/conjure/client/fennel/stdio.fnl
+++ b/fnl/conjure/client/fennel/stdio.fnl
@@ -99,6 +99,7 @@
       (stdio.start
         {:prompt-pattern (cfg [:prompt-pattern])
          :cmd (cfg [:command])
+         :env ["INPUTRC=/dev/null"]
 
          :on-success
          (fn []
--- a/doc/conjure-client-fennel-stdio.txt
+++ b/doc/conjure-client-fennel-stdio.txt
@@ -6,7 +6,6 @@ CONTENTS                                *conjure-client-fennel-stdio-contents*
     1. Introduction ........ |conjure-client-fennel-stdio-introduction|
     2. Mappings ................ |conjure-client-fennel-stdio-mappings|
     3. Configuration ...... |conjure-client-fennel-stdio-configuration|
-    4. Bugs ........................ |conjure-client-fennel-stdio-bugs|
 
 ==============================================================================
 INTRODUCTION                        *conjure-client-fennel-stdio-introduction*
@@ -68,17 +67,5 @@ All configuration can be set as described in |conjure-configuration|.
             Command used to start the Fennel REPL, you can modify this to add
             arguments or change the command entirely.
             Default: `"fennel"`
-
-==============================================================================
-BUGS                                        *conjure-client-fennel-stdio-bugs*
-
-If you find that after enabling the fennel stdio client, Neovim becomes
-unresponsive and/or generally broken when opening .fnl files, the fennel
-readline integration is probably breaking things. In this case, you should
-uninstall the lua readline module, remove it from your `LUA_PATH` environment
-variable or use an alternate lua interpreter by setting the command variable.
-For example, with this option:
->
-  let g:conjure#client#fennel#stdio#command = "fennel --lua /usr/bin/lua5.4"
 
 vim:tw=78:sw=2:ts=2:ft=help:norl:et:listchars=
--- a/fnl/conjure/remote/stdio.fnl
+++ b/fnl/conjure/remote/stdio.fnl
@@ -98,7 +98,8 @@
 
     (let [{: cmd : args} (parse-cmd opts.cmd)
           (handle pid) (uv.spawn cmd {:stdio [stdin stdout stderr]
-                                      :args args}
+                                      :args args
+                                      :env opts.env}
                                  (client.schedule-wrap on-exit))]
       (stdout:read_start (client.schedule-wrap on-stdout))
       (stderr:read_start (client.schedule-wrap on-stderr))
--- a/lua/conjure/client/fennel/stdio.lua
+++ b/lua/conjure/client/fennel/stdio.lua
@@ -275,7 +275,7 @@ do
         local function _5_()
           return display_repl_status("started")
         end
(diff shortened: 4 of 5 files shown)


fnl/conjure/client/fennel/stdio.fnl
```diff
@@ -99,6 +99,7 @@
       (stdio.start
         {:prompt-pattern (cfg [:prompt-pattern])
          :cmd (cfg [:command])
+         :env ["INPUTRC=/dev/null"]
```
Olical, line 102: Oh, it takes a sequential table of strings in this format? What an odd API :sweat_smile: part of me feels like this should be a associative table of keys to values that we convert into this format, i.e: `{:INPUTRC "/dev/null"}`. This allows easy overriding or removal of keys without hunting through strings for matches. What do you think?

Also (once we're sure on the format) it'd be great if you could add the entry to the `fennel-stdio` help text, under the one for the `cmd` config (should just be copy/paste/edit a couple of lines).

nicopap: Actually finding the proper format is what took me the most time. The neovim docs links to the luv doc, which just says there is an `env` option to define the environment. So I looked into the c implementation of luv, which was not very helpful. It seemed to use ipair to iterate the `env` option and set strings in an array it passes to libuv. It's only looking up [some doc related to the C libuv API](https://nikhilm.github.io/uvbook/processes.html#spawning-child-processes) that I figured it out. For the explanation: The format looks like the C posix `environ` (`man 3p exec`). In fact, [I can find the exact line of code where our weirdly formatted fennel string gets assigned to `environ`](https://github.com/libuv/libuv/blob/217fdf4265589889d00c7c0622fde2710971a020/src/unix/process.c#L306)!

Regarding your suggestion. I think it's a good idea. After all we ain't limited by C here. I'll add that and the doc amendment tomorrow.

Olical: Oh wow, yeah, that's frustrating! Well done for working it out, not sure if I would've got to the bottom of it. I've found myself looking through C for libuv things too, pretty awful state of documentation from the Lua perspective anyway :sweat_smile:

I'm about right now, so I may well just merge and handle the doc / tableification myself, saves you fiddling with it :slightly_smiling_face:

Olical:
I now realise you aren't exposing this as an option, so it's totally fine as it is. I will merge and adjust it slightly so that all stdio clients benefit from this since it's not just limited to Fennel really. Thank you very much!

Olical:
Hmm it doesn't seem to work for me :thinking: the fennel stdio client doesn't start up the REPL (no output) if I have `:env` set... will try to debug!

Olical:
Fixed it I think, see https://github.com/Olical/conjure/commit/4eab13e0a60e047db7f87376a1fcb810ca8dbd1b

Not sure why it was required for me? Maybe NixOS? I _need_ to explicitly inherit from the parent env, which is kinda okay but weird that there's no option in libuv to say "override and inherit" or "replace". I think this approach is okay? The default (`NULL` :env`) uses the parent env, setting it to anything totally replaces it.

Olical:
Will wait to confirm you're happy then probably do the release dance :smile:

nicopap:
Yeah. By default it's inheriting the nvim environment and if `env` is set, it completely overrides it, so you need to merge with the current environment. Apparently it didn't break anything for me. Glad you caught it. I'm sorry, that could have been a nasty bug. Your improvement is alright for me.

PR #175, opened 2021-02-14
Add a stdio based Hy client

peaceamongworlds:
This is a conjure client for Hy. It's essentially a copy of the Racket client, but with a few changes to make it compatible with Hy.

The client is currently quite basic, but all of the essential features work. I would like to add pretty-printing to the output of the Hy REPL, but this doesn't seem to quite work right now. As it is stdio based, all that it requires is that hy is installed on the system.

Please tell me if there's anything that I need to change or add.


Olical:
Oh this is neat! Thank you! And yeah, these copy and paste jobs are expected and welcome :slightly_smiling_face: I deliberately go for some level of duplication to keep things flexible enough to work with different languages and paradigms.

The only thing I'd change is the right alignment of the `*headings*` in the help text, I think they're like two characters from the right hand of the `=====` separator? It's a real nitpick of course.

I'll give it a go locally and add some review comments for small things, the general idea is great, thank you :smile:

fnl/conjure/client/hy/stdio.fnl
```diff
@@ -0,0 +1,168 @@
...
+      (display-repl-status :stopped)
+      (a.assoc (state) :repl nil))))
+
+; (defn enter []
+;   (let [repl (state :repl)
+;         path (nvim.fn.expand "%:p")]
+;     (when (and repl (not (log.log-buf? path)))
+;       (repl.send
+;         (prep-code (.. ",enter " path))
+;         (fn [])))))
```
Olical, line 112: Safe to delete :fire:

fnl/conjure/client/hy/stdio.fnl
```diff
@@ -0,0 +1,168 @@
...
+           (stop))
+
+         :on-stray-output
+         (fn [msg]
+           (display-result msg))}))))
+
+(defn on-load []
+  ; (augroup
+  ;   conjure-hy-stdio-bufenter
+  ;   (autocmd :BufEnter (.. :* buf-suffix) (viml->fn :enter)))
```
Olical, line 154: Also deletable :fire:

Olical (changes requested):
Just small things, mostly deleting unused comments :sweat_smile:

Without module naming in the file I guess it'll be hard to work with private things, but this is better than nothing! Maybe one day we can add an nREPL client alongside this one.

Oh! One more thing I thought of, adding it to the README and a wiki entry like this https://github.com/Olical/conjure/wiki/Quick-start:-Fennel-(stdio) which I'm happy to do really. Just small things I add when adding a new client.

fnl/conjure/client/hy/stdio.fnl
```diff
@@ -0,0 +1,168 @@
...
+(config.merge
+  {:client
+   {:hy
+    {:stdio
+     {:mapping {:start "cs"
+                :stop "cS"
+                :interrupt "ei"}
+      :command "hy --repl-output-fn=hy.contrib.hy-repr.hy-repr"
+      ; :command "hy --repl-output-fn=hy.contrib.pprint.pprint"
+      :prompt-pattern "=> "}}}})
```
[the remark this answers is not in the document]
peaceamongworlds: I haven't found any cases of where this doesn't work, but using `\n=> ` always fails.

Even evaluation forms like `(setv foo=> 5)` and `foo=>bar` works just fine.

I think if there are cases where this fails, then hy has an option to change the default prompt, so it can just be set to a uuid to guarantee that there won't be a clash, however I think that there might still be an issue with the very first prompt in the repl.

peaceamongworlds:
> Just small things, mostly deleting unused comments sweat_smile

I've cleaned up my code a little, so hopefully that should all be good now. I also removed the batched processing of messages. I think it makes long running jobs quite confusing as you don't get any output until the command has finished running. It makes running the callbacks for the HUD and answer substitution a bit tricky, but there isn't a perfect way of doing it since there's not way to tell if an output is just something being printed to stdout or the actual value that's being returned.

> Without module naming in the file I guess it'll be hard to work with private things, but this is better than nothing! Maybe one day we can add an nREPL client alongside this one.

I'd want to get an nREPL client, but the current implementation of an nREPL server in hy is just broken unfortunately.

> Oh! One more thing I thought of, adding it to the README and a wiki entry like this https://github.com/Olical/conjure/wiki/Quick-start:-Fennel-(stdio) which I'm happy to do really. Just small things I add when adding a new client.

I don't know how to add to the wiki. Could you do this yourself? The quick start guide should be really short: as long as you have hy installed on your system, everything should just work.

Olical:
Merged and added the wiki entry, thanks for taking the initiative. The batching of messages part is maybe okay, although it breaks some of my assumptions I think?

So this Hy client might not work correctly with eval and replace (`<prefix>e!`), inline eval results and others that try to get the result. I think the Racket one does, so there's a trade off. Delaying messages slightly until the prompt reappears means we know roughly what the result is.

Something isn't right about the current implementation as far as I can tell, but I'll give it some more thought. If you can improve it and get it working consistently with `<prefix>e!` then that'll be amazing, but I'll have a go too I think. Regardless, getting a first version of this out is cool, improving it into the future is always possible :smile: we might be able to replace it with an nREPL like client some day anyway.

peaceamongworlds:
> Merged and added the wiki entry, thanks for taking the initiative. The batching of messages part is maybe okay, although it breaks some of my assumptions I think?
> 
> So this Hy client might not work correctly with eval and replace (`<prefix>e!`), inline eval results and others that try to get the result. I think the Racket one does, so there's a trade off. Delaying messages slightly until the prompt reappears means we know roughly what the result is.
> 
> Something isn't right about the current implementation as far as I can tell, but I'll give it some more thought. If you can improve it and get it working consistently with `<prefix>e!` then that'll be amazing, but I'll have a go too I think. Regardless, getting a first version of this out is cool, improving it into the future is always possible smile we might be able to replace it with an nREPL like client some day anyway.

I'm currently keeping track of the last received value before the command is finished, and then assuming that that value is the output of the command, then only calling the callback function on the last value. This sort of allows the eval and replace and inline results to work. However it's obviously not perfect as I don't know if the last value is the actual output or just something printed to stdout.

Personally I'd rather have long running commands work correctly. Especially with hy, eval and replace seems like a bit of an esoteric feature.

The problem seems to stem from the fact that the hy repl doesn't produce any output when the returned value is `None`, hence there is no way to be certain that the last captured value is the real return value. Using something like `*1` to get the last value also doesn't work because it updates the all the other `*i` variables. I think that this will require a change in the hy repl itself.

PR #178, opened 2021-02-20
Guile client via Unix domain socket (guile --listen)

taw10:
This implements a new client, conjure.client.guile.socket, which connects to a running Guile process via a Unix domain socket.  eval-file and doc-str are not implemented yet, but everything else works including basic module switching.  Connecting to a network socket should be an easy future extension.

There is a similar problem to #166 with Guile.  The user setup file (~/.guile) can be skipped with "guile -q", but unfortunately that choice applies to both the socket REPL used by Conjure and the "normal" REPL which might also be in use.  I couldn't find any simple way to de-activate the readline/coloured prompt, once activated, for just one of the REPLs.  Therefore, I went the route of stripping out terminal control characters.  See function strip-unprintable in conjure.remote.socket, which works fine at least for the "default" readline/colorized Guile setup.  I don't think many people customise their Guile prompts beyond that - but maybe I'm wrong!

The generic part (conjure.remote.socket) is based on conjure.remote.stdio.  The stdio prompt parsing was not sufficient for Guile, so I added a way to provide a function for parsing the prompt, including detecting error conditions.  Guile's output looks like this:
```
scheme@(guile-user)> (+ 1 1)
$19 = 2
scheme@(guile-user)> (display "Hello, world!\n")
Hello, world!
scheme@(guile-user)> (begin (display "Hi there\n") (+ 1 1))
Hi there
$20 = 2
scheme@(guile-user)>
```
There is therefore some opportunity to (unreliably) distinguish between return values and other output.  An error condition looks like this:
```
scheme@(guile-user)> (+ 3 ff)
;;; <stdin>:29:0: warning: possibly unbound variable `ff'
ice-9/boot-9.scm:1669:16: In procedure raise-exception:
Unbound variable: ff

Entering a new prompt.  Type `,bt' for a backtrace or `,q' to continue.
scheme@(guile-user) [1]> ,q
scheme@(guile-user)>
```
... the Conjure client immediately sends ',q' in response to seeing the error (detected by the extra [1] in the prompt).

A 'stdio' client, where Conjure starts the Guile process, would also be possible.  It would require the stdio remote to allow a prompt-parsing function (vs just one pattern, as now).  In your opinion, given that there is at least some structure to the output, would such a client still count as 'conjure.client.*.stdio'?  Would it be worth extending the stdio remote to enable this, or would you prefer to keep the stdio remote as simple as possible?  Starting Guile separately is more flexible and useful, but having Conjure start the Guile process might allow people to get up and running quicker.

I would like to implement eval-file and doc-str before this gets merged, but maybe you have some comments on the code so far.


taw10:
Update: eval-file and doc-str now implemented!

Olical:
Oh wow, this looks great! Excellent attention to detail and splitting of modules, you nailed it :D only feedback so far (just had an initial skim) is that it's pointing at `master` when I'll be merging this into `develop` first. Should be an easy fix and compatible.

Olical:
There's `strip-ansi-escape-sequences` in `conjure.text` already which might be some duplication with your control code stripping?

I think the new remote module is cool and I'd probably try to keep stdio generic but maybe add some extra tools to make it work better for guile.stdio. And having both options is possible by design, defaulting to the canonical /  best implementation (probably socket?) then having some config to switch to the other kind is expected. Fennel switches between the built in Aniseed client and the fennel.stdio one.

I think having different transports/remotes which are generic and simple but open to enough extension to support every client is the way to go. `stdio` is about running under the nvim process, `socket` would be a similar text interface but over socket imo. Both could have similar "parsing" hooks, although that parsing should hopefully be doable in the client itself once the remote passes the data to it in the callback.

taw10:
Ok, I've switched the base branch to develop.  Sorry about the mess!  If you want to avoid an extra merge in the history, I'm happy to re-base.
I tested using strip-ansi-escape-sequences.  It gets rid of the colour sequences, but leaves some \1 and \2 characters in.  I'm not sure what they're doing in the prompt.  I replaced the inner part of my control-sequence splitting with strip-ansi-escape-sequences to avoid the duplication.

Olical:
Nice thanks! And I don't care about that, no worries :slightly_smiling_face:

And nice one on the dedupe! Maybe it'd be worth extending the one shared one at some point, but I'm okay with this. Time for a mergeroo and testing / soaking on develop. I'll also set it up in the config with my new dotted filetype support.

Olical:
Thank you so much for this work! If you have more you want to do, fire away, I'll just get it merged into develop so it can soak a little and is easier for others to try.

Olical:
I might make some amendments to the UX :thinking: especially around warning you about what's not set etc. See if I can get it working a little more like the other clients while still keeping it convenient for you. Might not be merging for a little while though, quite a lot on right now :sweat_smile:

Olical:
Currently set it up to launch for scheme _or_ guile _or_ scheme.guile filetypes too.

PR #179, opened 2021-02-21
remote.socket: Use strip-ansi-escape-sequences instead of gsub

taw10:
Sorry, I'd make a classic "forgot to push" mistake with #178.  Here is the missing de-duplication!


Olical:
Haha, no problem!

PR #187, opened 2021-03-04
bare bones support for mit-scheme

callum-oakley:
Opening this as a draft because a lot still doesn't work, but the absolute minimum requirements of evaluating forms and seeing the results are working fairly consistently, which is already enough for it to be useful for me.


fnl/conjure/client/mit-scheme/stdio.fnl
```diff
@@ -0,0 +1,136 @@
+;;;; Known issues, non-exhaustive:
+;;;;
+;;;; TODO Exiting vim without first stopping the REPL with <leader>cS causes
+;;;; mit-scheme to spike to 100% CPU and stay there until it's killed.
+;;;;
+;;;; TODO Evaluate word doesn't work. After trying to evaluate word I get no
+;;;; output at all until I stop the REPL at which point I get all the output at
+;;;; once.
```
Olical, line 8: Probably need to append a newline to every chunk of code you're sending to eval.

fnl/conjure/client/mit-scheme/stdio.fnl
```diff
@@ -0,0 +1,136 @@
+;;;; Known issues, non-exhaustive:
+;;;;
+;;;; TODO Exiting vim without first stopping the REPL with <leader>cS causes
+;;;; mit-scheme to spike to 100% CPU and stay there until it's killed.
```
Olical, line 4: I'll hopefully fix this in a generic-ish way soon. I want to provide a hook that gets called on Neovim exit that you should use to stop the sub-process.

fnl/conjure/config.fnl
```diff
@@ -57,7 +57,10 @@
               :clojure :conjure.client.clojure.nrepl
               :janet :conjure.client.janet.netrepl
               :hy :conjure.client.hy.stdio
-              :scheme :conjure.client.guile.socket
+              ;; TODO leave this as is, make mit-scheme an optional alternative
+              ;; to guile
+              ; :scheme :conjure.client.guile.socket
```
Olical, line 62: Yeah, not sure on the defaults right now... it'll be arguable either way as to which Scheme is the more default scheme I guess! Maybe MIT is more canonical? I think picking one as a default that works okay (maybe MIT tbh since it's a sub process) then documenting it and making it easy to switch is the way to go.

I _think_ guile gets `scheme.guile` as it's filetype anyway, so if you have the vim-guile plugin just configuring `guile` with Conjure will be enough for it to match up nicely. Conjure looks through all dotted parts of the filetype to find a matching client, so `guile` in Conjure will match `scheme.guile`.

Olical (commented):
This is looking really good so far! Well done! Hopefully I can add some stuff to address the clean shutdown behaviour for you soon. I'll leave this as an open PR for as long as I can until I get around to adding my own touches :smile: the TODOs are helpful, maybe I'll end up addressing most but if you beat me to it that's great too!

Olical:
Merging so I can tinker :slightly_smiling_face:

Issue #113, opened 2020-08-16
Feature Request: Highlight evaluation


davidsierradz:
Hi, I'm enjoying playing with Conjure!

I was trying to add a brief highlight when evaluating an expression, I hacked something in Lua (My fennel and Lua knowledge is almost non-existent) 

It looks something like this. Here I'm evaluating this root-form:

![image](https://user-images.githubusercontent.com/9190258/90339741-72218f80-dfb8-11ea-95cf-83606216b847.png)

Or check this asciinema:

[![asciicast](https://asciinema.org/a/353693.svg)](https://asciinema.org/a/353693)

I got:

```patch
diff --git a/lua/conjure/eval.lua b/lua/conjure/eval.lua
index ed21063..f27b746 100644
--- a/lua/conjure/eval.lua
+++ b/lua/conjure/eval.lua
@@ -128,11 +128,35 @@ do
   client_exec_fn = v_0_
 end
 local eval_str = nil
+local function highlight_range(opts)
+  if not (type(opts.start) == "table" and type(opts.finish) == "table") then
+    print("arguments start or finish: expected table")
+    return nil
+  end
+  local start = opts.start
+  local finish = opts.finish
+  local start1 = {}
+  start1[1] = start[1] - 1
+  start1[2] = start[2]
+  local finish1 = {}
+  finish1[1] = finish[1] - 1
+  finish1[2] = finish[2] + 1
+  local bufnr = opts.bufnr or 0
+  local higroup = opts.higroup or "IncSearch"
+  local timeout = opts.timeout or 500
+  local NAMESPACE = vim.api.nvim_create_namespace "conjure_highlight"
+  vim.highlight.range(bufnr, NAMESPACE, higroup, start1, finish1)
+  vim.defer_fn(
+    function() vim.api.nvim_buf_clear_namespace(bufnr, NAMESPACE, 0, -1) end,
+    timeout
+  )
+end
 do
   local v_0_ = nil
   do
     local v_0_0 = nil
     local function eval_str0(opts)
+      highlight_range({start=opts.range["start"], finish=opts.range["end"]})
       local function _3_()
         if opts["passive?"] then
           return opts
```

The function accepts a range (Conjure range and Neovim range have an offset, nothing difficult), a buffer number, a highlight group and a duration.

Is this something you think would be valuable in Conjure (Maybe behind a config flag)? The problem is: I don't know how to write the fennel code that outputs this Lua.

Thanks!

Olical:
[...]

PR #193, opened 2021-03-14
Implement highlight evaluation

taw10:
This implements highlights for evaluation as described in #113.  I did an almost direct conversion from @davidsierradz's Lua code to Fennel.

To configure (hopefully self-explanatory):
```
let g:conjure#highlight#enabled = 1
let g:conjure#highlight#timeout = 500
let g:conjure#highlight#group = "IncSearch"
```
These are the defaults, except for `enabled` which defaults to false.

Unfortunately, this requires a very recent NeoVim to work: more than 0.4.4, which is the latest release version at the time of writing.  I chose to make it silently fail on older versions.

I'm not sure if I called the Vim API functions in the correct idiomatic way.  I struggled to use `nvim` (i.e. `conjure.aniseed.nvim`) - it seemed to choke on the function name containing a dot (`highlight.range`).  Perhaps I didn't try hard enough.

**Question/rationale for defaulting to disabled:** Is there a way to set an option to `false` with normal `.vimrc`?  Setting a `g:conjure#something` variable to `0` in fact enables the option.  `debug` seems to work the same way, so I didn't do anything extra to make the behaviour more logical (to me).

### DIFF
--- a/fnl/conjure/eval.fnl
+++ b/fnl/conjure/eval.fnl
@@ -28,6 +28,32 @@
     [opts.preview]
     (a.merge opts {:break? true})))
 
+(defn- highlight-range [range]
+  (when (and (config.get-in [:highlight :enabled])
+             vim.highlight)
+    (let [bufnr (or (. range :bufnr) 0)
+          namespace (vim.api.nvim_create_namespace "conjure_highlight")
+          hl_start {1 (- (. range.start 1) 1)
+                    2 (. range.start 2)}
+          hl_end {1 (- (. range.end 1) 1)
+                  2 (. range.end 2)}]
+
+      (vim.highlight.range bufnr
+                           namespace
+                           (config.get-in [:highlight :group])
+                           hl_start
+                           hl_end
+                           :v
+                           true)
+
+      (vim.defer_fn
+        (fn []
+          (vim.api.nvim_buf_clear_namespace bufnr
+                                            namespace
+                                            0
+                                            -1))
+        (config.get-in [:highlight :timeout])))))
+
 (defn- with-last-result-hook [opts]
   (let [buf (nvim.win_get_buf 0)
         line (a.dec (a.first (nvim.win_get_cursor 0)))]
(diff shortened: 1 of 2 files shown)


Olical:
Awesome! Thank you! I'll take a look through it soon but wrt your questions:

 * dotted `nvim.fn.*` calls: That may be a bug in https://github.com/norcalli/nvim.lua (which is all `conjure.aniseed.nvim` is at this point), I'll have a look though! I intend to add my own nvim lib to aniseed based on Fennel macros some day, so maybe I can "fix" it by doing that eventually.
 * False in viml: You can use `v:false`, I think it's the best way to express boolean values. There's also `v:true` and `v:null` I think?

harrygallagher4:
Regarding calling `vim.highlight.range`, [nvim.lua][] doesn't wrap the nvim lua "standard library" functions because they're already written as lua functions/libraries. A lot of functions from [nvim.lua][] have actually been added to nvim itself so it seems like all it's really needed for is shortening calls to `vim.api.*` and `vim.cmd`. Basically any function call that doesn't start with `vim.api.nvim_` is not available in the wrapper. I would recommend reading `:help lua` and `:help api` to understand the difference if I've explained it poorly.

@Olical I've also considered writing some nvim fennel macros too, did you have any specific ideas in mind? Perhaps I could work on contributing those to aniseed. Oh yeah, by the way, thanks so much for aniseed & conjure I absolutely love both of them!

```lua
-- builtin
vim.api.nvim_buf_set_lines(...)
-- wrapper
nvim.buf_set_lines(...)

-- builtin
vim.cmd('echo ...')
-- wrapper
vim.ex.echo(...)

-- not part of wrapper
vim.highlight.range(...)
vim.inspect(...)
vim.loop.new_timer()
vim.regex()
```

[nvim.lua]: https://github.com/norcalli/nvim.lua

fnl/conjure/eval.fnl
```diff
@@ -28,6 +28,32 @@
     [opts.preview]
     (a.merge opts {:break? true})))
 
+(defn- highlight-range [range]
+  (when (and (config.get-in [:highlight :enabled])
+             vim.highlight)
+    (let [bufnr (or (. range :bufnr) 0)
```
Olical, line 34: Instead of `or` to a `0` here we should look up the current buffer number since the user could leave the current buffer during the timeout, causing the highlight to never actually be cleared.

(I'll do this, just making a note for myself)

PR #208, opened 2021-05-24
Add help tags for commands and configuration options

nbardiuk:
I find help tags for commands and configuration options very useful:

* easier to point somebody to a specific option like `:help g:conjure#log#wrap`
* allows jumping to documentation using `K` right from usage in init.vim
* allows jumping to mentions inside documentation using `K` or `Ctrl-]`
* makes options and commands more discoverable using `Helptags` from fzf-vim, vim-clap, or telescope

### DIFF
--- a/doc/conjure-client-clojure-nrepl.txt
+++ b/doc/conjure-client-clojure-nrepl.txt
@@ -45,6 +45,7 @@ options relevant to these mappings.
                          `:ConjureConnect 5678`
                          `:ConjureConnect`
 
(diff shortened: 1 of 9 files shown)


Olical:
Ohh this is great, thank you so much! I wasn't sure if we could do this by wrapping the existing code on the left hand side or not. Let's go with this for now and maybe we can find a way one day to consolidate the tags and the names of the config into one place. So the example like `g:conjure#debug` is also the tag.

Thanks again!

PR #222, opened 2021-07-10
HUD: Add border option

p00f:
image


p00f:
The border looks thin only because of my colorscheme

PR #233, opened 2021-08-05
Add a saner default for non-Lisps

glyh:
I've added some configs so that default config for lisp and non-lisp are different, which is often the way we want. 

This makes seamless interop with [nvim-treesitter/nvim-treesitter-textobjects](https://github.com/nvim-treesitter/nvim-treesitter-textobjects).


fnl/conjure/config.fnl
```diff
@@ -77,6 +82,8 @@
     :log_reset_soft "lr"
     :log_reset_hard "lR"
 
+    ; These get abadoned if using non_lisp config.
+    ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
Olical, line 86: Hmm why would the current form not make sense in a non-lisp?

```js
if (foo)
  bar()
} else {
  baz()
}
```

In that JS snippet I could consider the following square bracket denoted forms:

```js
[if ([foo])
  [bar()]
} else {
  [baz()]
}]
```

glyh: Though I have to say, that there still need some improvement on treesitter-textobjects, they are not enough textobjects that lispers may really want(the textobjects over there are mostly not expression-based)

glyh:
I've found that the non expression based languages may actually requires a new different type of evaluation, so maybe add a new command, evaluate statements to it maybe better.

glyh:
I may close and reopen this with some other modifications, how do you think about evaluation for non-lisps?

Olical:
Hmm I'm not sure if I'm on board with the idea of splitting into "lisp" and "non-lisp" unless it's absolutely the very last resort. Any split adds potentially double maintenance cost and worry as I break one and not another without realising. So I try to avoid special casing outside of the clients as much as possible.

I'd rather add hooks into the client interface that allows the individual clients to add their own selection logic that isn't based on well defined form indicators (like parens in lisps). So a hook you can define for a client that returns true if what you have can be evaluated as an expression or if the TS code should continue to walk further up.

That's just one idea off the top of my head, but I'd much rather extend the client design / interface than add custom things outside of that in the shared layer since I want that to remain as generic as possible, I don't want the outer shell of Conjure to know anything about the languages it's talking to, that's the point in the client interfacer really (which is only defined at the bottom of `:h conjure` which isn't great, I should codify it so it can be verified programmatically).

glyh:
Well, I'll rework and you can see if it meets your requirements

PR #241, opened 2021-08-19
Add `conjure` help tag definition

stefanvanburen:
Hi,

Might be just me, but typically when I'm looking for help for the plugin, I like to go to the main help ... file? for that plugin and then proceed to search from there. (At least, this is how [vim-go](https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt#L2) seems to do things).

With this change, when I type `:help conj` followed by <kbd>Tab</kbd>, I get `conjure` as the top result.

Anyway, no worries if you don't want to merge this, I can always just be a little more picky in my `:help conj` searching - thanks for the plugin!


Olical:
Hey, this is fine by me, thanks!

PR #274, opened 2021-11-28
Run tests will respect raw-out setting

qnkhuat:
1. For all run tests command in Clojure, the output now will respect the `g:conjure#client#clojure#nrepl#eval#raw_out` setting.
This makes debugging easier IMO.

Examples :
without raw out
Screen Shot 2021-11-28 at 22 28 19
with raw out
Screen Shot 2021-11-28 at 22 27 55

2. I also remove the log-append `Success!` if the test is successful because I find logging `nil` is more familiar since it's the default output of `clojure.test/test-vars`.

### DIFF
--- a/fnl/conjure/client/clojure/nrepl/action.fnl
+++ b/fnl/conjure/client/clojure/nrepl/action.fnl
@@ -455,7 +455,8 @@
         #(ui.display-result
            $1
            {:simple-out? true
-            :ignore-nil? true})))))
+            :ignore-nil? true
+            :raw-out? (cfg [:eval :raw_out])})))))
 
 (defn- run-ns-tests [ns]
   (try-ensure-conn
@@ -469,7 +470,8 @@
           #(ui.display-result
              $1
              {:simple-out? true
-              :ignore-nil? true}))))))
+              :ignore-nil? true
+              :raw-out? (cfg [:eval :raw_out])}))))))
 
 (defn run-current-ns-tests []
   (run-ns-tests (extract.context)))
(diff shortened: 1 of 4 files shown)


fnl/conjure/client/clojure/nrepl/action.fnl
```diff
@@ -514,16 +516,11 @@
...
                  :context (extract.context)}
-                (nrepl.with-all-msgs-fn
-                  (fn [msgs]
-                    (if (and (= 2 (a.count msgs))
-                             (= "nil" (a.get (a.first msgs) :value)))
-                      (log.append ["; Success!"])
-                      (a.run! #(ui.display-result
-                                 $1
-                                 {:simple-out? true
-                                  :ignore-nil? true})
```
Olical, line 525: Do you not want to just pass `raw-out?` here? Maybe you accidentally raised the display-result form? I think that's probably what happened here?

Olical:
So I've taken this idea and pushed it into develop in a slightly different way. I've made a separate raw_out flag so there's `g:conjure#client#clojure#nrepl#eval#raw_out` and `g:conjure#client#clojure#nrepl#test#raw_out`. You can flip them independently so you can have one on and not the other.

I've also made this change in a way that doesn't remove the forms I pointed out in your PR, I think you accidentally deleted some code so I've made the change without doing that.

Let me know what you think! It's on the `develop` branch 😄

Olical:
Closing this since we're making the same changes now near enough.

qnkhuat:
Hi @Olical, I'm trying to try out the new version you pushed on `develop`.

1. I've tried to set `let g:conjure#client#clojure#nrepl#test#raw_out=1`, but my output is still being covered in comments. Could you take a picture of how your output looks like when you apply that flag?
   - note that I installed conjure using Plug `Plug 'Olical/conjure', {'branch': 'develop', 'for': ['clojure']}`
   
2. I saw that we still set `ignore-nil` to true. Please consider setting it to `true`, because currently, it's hard to know when the tests are finished when running a single test. By setting it to true we always have a `nil` message when the tests are finished in all cases.

3. When tests succeed but produce output during its execution(print or log), it will not log `Success`. this sometimes creates confusion because Conjure doesn't behave consistently.

Thanks!!!

Olical:
What happens if you set it to `v:true` instead of `1`?

qnkhuat:
Just tried it, still not works :/

PR #278, opened 2021-12-08
try to parse host and port from single argument to `:ClojureConnect`

kevinkrouse:
Small change to parse a host and port if there is a single argument to `:ClojureConnect`


Olical:
Neat! Will take a look at this ASAP, thank you!

Issue #279, opened 2021-12-09
Go to definition on jar files is not working on nvim 0.6


rafaeldelboni:
Looks like some change in the new version on `pi_zip v32` on Oct 22, 2021 changed the path for zipfiles and when I try to do a go to definition it opens a blank buffer

- https://neovim.io/doc/user/pi_zip.html
- https://github.com/tpope/vim-fireplace/pull/398

Olical:
More breaking changes!? 😭 but thank you so much for keeping an eye on this and fixing it. I'll merge now and try out the changes today then deploy ASAP, I've got a super busy weekend ahead but at least it'll be on the develop branch.

Thank you! ❤️

Olical:
Oh no, it was targeting master, I need to fiddle a bit

rafaeldelboni:
Oh I'm sorry I forgot about the develop branch :(

As I suggestion, if you add a note about the branch on CONTRIBUTING.adoc, which btw helped me a lot :)

Olical:
Not a problem! I'll do that, for now I've just merged it into develop, will leave it on both so it's technically stealth release at the moment but I've got a lot of work to do so that'll be fine for now.

PR #280, opened 2021-12-09
[Clojure] fix: Bug in go to definition on jar/depencies using v32 pi_zip

rafaeldelboni:
Fixes: https://github.com/Olical/conjure/issues/279


rafaeldelboni:
Tested on v31 and v32 and is working fine! :)

PR #282, opened 2021-12-14
add highlight and prefix options for inline results

p00f:
the doc for `conjure#eval#inline#prefix` looks wrong but I'm not sure what to put there


p00f:
Should I also change `conjure#eval#inline#results` to `conjure#eval#inline#enable`?

Olical:
Ohh this is a neat change! 👀  thank you!

Olical:
As far as the naming goes, I know it's not perfect and really it should all be grouped under one table but I'm okay with it as it is. It'll require a migration path / time to change it and I honestly don't think it's worth it right now. It's already under the eval grouping which keeps things neat enough for my liking.

I also wrapped some values in the docs in quotes which improves the syntax and hints that you want to set it to strings.

Also removed a `(local...)`, just because we should either really use `let` (in my opinion) or just inline things if they're used once. and aren't that complex. I ended up removing the local in favour of inlining.

Merging! Thank you very much!

Olical:
it's merged, GitHub is just very confused!

PR #288, opened 2022-01-02
Support for inspecting tree-sitter node type

glyh:
As title. Currently client don't know the type of the node. With this PR, this can be resolved, and later we can add support for non-lisps base on this.


glyh:
Nevermind, don't accept this. I found that more tree-sitters side information should be exposed for this to be useful.

glyh:
I've found out that client side can directly do some interact with nvim-treesitter, so this might be of no use.

Olical:
Oh interesting, yeah shouldn't be an essential part of Conjure really. Maybe internally to ask "is this a string or a string's contents? Because I want the whole string!" but that's all. Could be a good plugin to build standalone though!

PR #289, opened 2022-01-02
Fix Cword Range

glyh:
This PR fix the bug in word extraction from extract.fnl.


glyh:
Sorry but this is from master branch, I'll reopen this

PR #290, opened 2022-01-02
Fix cword range

glyh:
This PR fixs the bug in word function of extract.fnl, it use the correct range.

### DIFF
--- a/fnl/conjure/extract.fnl
+++ b/fnl/conjure/extract.fnl
@@ -122,12 +122,19 @@
         (a.last forms)
         (a.first forms)))))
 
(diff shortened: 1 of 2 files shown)


Olical:
Ohh this is interesting, thank you! That's been a very simple implementation for a long time and I agree it doesn't work great in some scenarios. I just got used to it but this is exciting.

Olical:
Oh I see what you've done! I thought you were fixing a bug with extracting the current word, I seed you finally implemented the word range! Thank you! LGTM, merging! 🎉

PR #291, opened 2022-01-02
Lua support

glyh:
This branch adds lua support via neovim internals.


glyh:
BTW, I found ts-support some what strange, for example, I evaluate current form for :
```lua
"abc"
```
when the cursor is on one of the `"`s, it would turn out to evaluate symbol `abc` rather than string `"abc"`

glyh:
Also, this PR now has some problem.  Though codes will be correctly evaluated and logged, the HUD does not always appear.

I would be happy if you can instruct me how to fix this.

Olical:
The HUD not appearing _can_ be a symptom of the cursor moving as you evaluate which means that these events happen:

 * Eval using a text object such as `aw`
 * Result
 * Displayed in HUD
 * Cursor moves to beginning of evaluated text object
 * Closes HUD automatically

Which basically happens in a single frame, a debounce / cooldown period on the HUD opening and closing could fix this but also maybe introduce some horrible timing based bugs 🤔 adding timers in nvim is almost like adding threads anywhere else, here be dragons! But maybe helpful dragons!

As for the `abc` part, that's the tree-sitter implementation I think, you're using TS right? I think we need to modify `conjure.tree-sitter` to be smarter about walking up the tree to find the right thing to eval, I think someone raised an issue for the _exact_ problem you described.

Thank you for this PR though, I'm going to take a look now, this'll be an amazing addition.

Olical:
Testing locally now, you're a machine 🚀

glyh:
... Sorry but I'll reopen this. I rename the typo in branch name at my side

Olical:
Hah okay then, I don't mind at all but it's fine if you want to do that 😄

PR #292, opened 2022-01-03
Lua support

glyh:
This PR adds support for Lua via neovim internals


Olical:
This looks familiar 👀

glyh:
Yes, I renamed the branch at my side and this cause the PR to close...

Olical:
Sorry I was being silly 😅 I'm playing with it now, thank you for creating this! Were you evaluating using tree-sitter to select forms or using visual / range evals?

glyh:
Both. But TS support seems not that good. The optimal way, I think, is to use [nvim-treesitter-textsubjects](https://github.com/RRethy/nvim-treesitter-textsubjects) to select, and call evaluation on selection.

glyh:
Also, lua is strict on local variables, which will be clear to nil from evaluation to evaluation. One have to use global ones to preserve values. This is the same behavior as lua's REPL, but maybe it can be fixed?

Olical:
Yep, I had the same issues with evaluating Fennel. I eventually wrapped up the Fennel REPL implementation in a continuation based stateful function that allows me to create one eval context per buffer. This maintains local contexts between buffers which makes logical sense to the user I think.

Something like that for this would be fantastic... maybe with continuations again but I'm not sure how to do it right now. I'll have a think, maybe this will be good enough for now then we can add buffer local contexts later. I also want to look at keeping the HUD up after doing a visual eval since that's annoyed me for a while.

glyh:
I make a list of what to improve here (but maybe some of them is not doable):
- Investigate the way tree-sitter select nodes, and fix them. (maybe via other PRs)
- Investigate the behavior of HUD so that it will appear after each evaluation. (maybe via other PRs)
- preserve locals across evaluation. (may not be doable)
- Non-blocking evaluation.

Olical:
I think tree-sitter and HUD improvements are definitely possible. Preserving locals requires us to be able to store contexts and reuse them between evals (I have high hopes for this being possible!) but the non-blocking won't be possible I don't think.

Since Neovim is single threaded and the Lua interpreter is embedded within Neovim we're relying on that single process. So any Lua work MUST block until it's done. Unless maybe we create some sort of libuv threads but that'll probably lead to very very very weird bugs and problems. Neovim was not designed for that so I wouldn't want to mess with that ideally.

glyh:
OK, I'll invest some time to the first 3 issues later.

Olical:
I'm fixing the HUD one now 🙂  I have a fix, just working on config and naming.

Olical:
Sorry my change to the config added a conflict but `make compile` will fix it, no big deal.

Olical:
Would you be okay with me merging this to develop and resolving the conflict? You can open subsequent PRs but I'd like to get this onto develop in it's current state so I can work on it too and get people looking at it / trying it 😄

Olical:
Cool, let's go! Thank you for this, it's really cool! The HUD thing should be fixed already on develop too 🙂  added a minimum lifetime to the HUD for passive closing on cursor move which should catch all cases of cursor movement after eval.

Olical:
Merged! ❤️

Olical:
I don't _think_ the eval with persistent locals will be possible because I spent about a month trying to do it before for Aniseed. Here's a quick attempt but it doesn't work, just remembered it's probably what I tried first ages ago.

```fennel
(defn lua-repl []
  (let [env {}]
    (setmetatable env {:__index _G})
    (fn [code]
      (let [f (assert (loadstring code "some-chunk-name"))]
        (setfenv f env)
        (f)))))

(let [r (lua-repl)]
  (r "local n = 10")
  (r "print(n)"))
```

`n` is `nil` in the second eval, so you can set the env of the eval but it won't mutate the env table for reuse. That's the part I could never work out.

I got it working for Aniseed by hacking around with the REPL in `fennel.lua`, that does a LOT of work to make this possible since it's almost an entire stand alone interpreter. If we can work out a way to introspect the env of a function call or eval we can pull out locals and preserve them for the next eval.

If we can't then this will just have to remain a caveat of this client until someone basically writes a Lua in Lua interpreter that works flawlessly and exposes the locals each eval defines 😅

glyh:
I don't agree. 
```lua
-- WARN: please keep this default to be true!!!
_G._PAUSE_COLLECT_LOCALS = true
function _G.collect_locals(_, _)
  if _G._PAUSE_COLLECT_LOCALS then return end
  _G._PAUSE_COLLECT_LOCALS = true
  _G._LOCALS = {}
  local i = 1
  while true do
      local n, v = debug.getlocal(2, i)
      if not n then break end
      _G._LOCALS[n] = v
      i = i + 1
  end
  _PAUSE_COLLECT_LOCALS = false
end
debug.sethook(collect_locals, 'r')

function _G.call_and_collect(f, ...)
    _G._PAUSE_COLLECT_LOCALS = false
    f(...)
    _G._PAUSE_COLLECT_LOCALS = true
    return _G._LOCALS
end

function f(a, b, c)
    local d = a
    local e = b
    local f = c
end

call_and_collect(f, "hello", "world", "!")
```

Olical:
Yeah I've experimented with this route for probably 40+ hours of tinkering, they all ended in either dead ends or "this half works and it's super hacky and scary". The best thing that ever worked for me was wrapping up the Fennel REPL in continuations, so I could continually feed code into a stateful loop that I could pause and continue (through continuations). It would spit out results through callbacks...

I think I've actually thought of a way to do that just now... using basically the code I wrote with some tweaks (there's a different function that reads chunks over time from a function) + a callback it has access to that it'll call with every eval result... it'd work just like the Fennel REPL.

It all hinges on that Lua eval function being able to evaluate each chunk on it's own or if it needs to read to the end of the string before it evaluates. I'm mostly writing this down for my own notes for some time soon when I try it 🙂 will report back in when I get some time but I'm off holiday and back at work now

Issue #97, opened 2020-06-06
Support for Common Lisp


smolck:
Hello! So after seeing #92 and starting to use common lisp, I'd really like to implement a client for common lisp myself (that I could PR here) because I love the dev experience with Conjure, but I have some questions about that which hopefully you wouldn't mind answering?

1. Would you accept a PR adding common lisp support? If not I can create a separate plugin for it, but if you'd merge it here I'd like to do that.
2. How would you go about implementing a common lisp client? My idea was to create a repl (say via `sbcl`) and then communicate back and forth with that. Would that work, or is there a different/better approach?

I know there are plugins like vim-slime and such, but most of these seem old or outdated and they also don't make use of Neovim-specific features like floating windows (those are Neovim-specific, right?), unlike Conjure. So I'd much prefer to have a common lisp client for Conjure than use one of those.

Olical:
Hello! This is really exciting, I've been hoping someone would show interest in adding a client for a while :smile:

## Q1

Absolutely, 100%. As long as it's in line with the other clients and works well / has a good UX I'm more than happy to add it. I'm okay with a PR or a plugin, whatever works for you really.

If you go down the PR route you get to lean on everything that's already set up in the Conjure repo with regards to Aniseed Fennel compiling. If you write your own plugin (presumably with Aniseed too) you have more control over the marketing / readme / docs / tooling side. What do you think you'd prefer?

Since this is the first time someone has approached me about adding a new client I'm not sure on the best course. I have a feeling just integrating everything directly into Conjure is the way to go for ease of use by users. I'm leaning towards a fork that is eventually merged in.

## Q2

So I try to find some sort of socket based transport for each language, like netrepl for Janet and nREPL for Clojure. This way the user can start the process with all of the env vars and magical things they require, they can even start it in docker! Then the client just uses some port file or known good default port to try and connect when you enter the filetype for the first time.

I think starting `sbcl` under Neovim and reading/writing to that process is doable and I'd use that as a last resort if there's no good known network REPL solution in that language's community. I want to encourage getting your project running with a dev hook in it, then you start Neovim on the side and it automatically connects.

If Neovim starts the process you then tie them together, it becomes "jack in" support pretty quickly which requires so much config for users to get things running just how they want them. I want jack in _one day_ but I'm putting it off as long as possible.

[...]

PR #296, opened 2022-01-06
Common Lisp Client

Dotrar:
Continuing from the work greg's branch, I've implemented a fairly rudimentary but reliable (i hope) common lisp client. 

### Edit: 10th Jan
STDOUT now uses comment prefix to show up in the log. picture in [this comment](https://github.com/Olical/conjure/pull/296#issuecomment-1008428451) is what it looks like.

### Old picture, not quite correct

I've tried to keep it atomic, compiled after every commit, etc.

I'm out on holidays at the moment, and I have to walk 1.5 kms to get reception, so I probably won't reply until monday, but feel free to have a look and poke as many holes in it as possible and I can get them fixed up next week.

What defines a "client" that is ready to use? I've just gone for eval / stdout / errors here.That should be enough to try and build upon in the future?


Dotrar:
this should complete #97 in my estimation. 

some questions:
- Do you want some docs to go with this? I was thinking of writing out "what does a client need" sort of thing, and basic api info, but not sure what should be doc'd. (i just copied things) - this might be better in another PR.
- How is stdout meant to be shown in the conjure log? I just used a string representation, which seems good enough.  (i didn't check escape values, whoops)
- Is there any plans for stdin ? I was thinking something like the vim input command could work for simple cases, but could be hard to do

rafaeldelboni:
Thanks a lot for this! 😀

fnl/conjure/config.fnl
```diff
@@ -48,18 +48,19 @@
     nil))
 
 (merge
-  {:debug false
+  {:debug true
```
rafaeldelboni, line 51: I think you need to toogle this back to false right?

fnl/conjure/remote/swank.fnl
```diff
@@ -0,0 +1,64 @@
...
+                      (opts.on-failure err)
+
+                      (do
+                        (conn.sock:read_start (client.schedule-wrap handle-message))
+                        (opts.on-success)))))})))
+
+  ; (send conn (or opts.name "Conjure"))
+  conn)
+
+; (def c (connect {:host "127.0.0.1"
```
rafaeldelboni, line 58: Maybe cleanup this comments?

fnl/conjure/config.fnl
Dotrar: Yep! slipped by from greg's work. I've fixed this in [07159d7856ac7f6d8eb3103d331d5d1e5d7b734c]

fnl/conjure/remote/swank.fnl
Dotrar: also fixed in [07159d7856ac7f6d8eb3103d331d5d1e5d7b734c]

Dotrar:
Added doc word and eval-file. 
image

Olical:
I've still never used common lisp in any capacity so I'm taking your word for it when it comes to UX and behaviour right now but this looks fantastic. Excellent work! I'll have a look through the code from a Fennel / maintainer point of view but I might end up "LGTM"-ing some of the more nuanced CL parts since it's not my area yet.

I'll try to at least get a Hello World eval working though! I'll resolve the conflicts too.

Olical:
And thank you @rafaeldelboni for swooping in with early reviews and 👀 !

Olical:
At the surface level you've nailed the general file structure and my rough architecture. I'm going to merge-aroo and try it out locally, if things look okay let's get people trying it out on the develop branch for a couple of days then time for a release. There's a lot on the develop branch I want to get out right now.

Olical:
Took me a while to get a swank REPL up but it works for me! Nice! All the base features are there, I'll just knock up a basic help .txt but if someone wants to create the quickstart wiki page that'd be amazing. I think someone who actually knows some good CL install and setup tips would be best placed to do that though.

Olical:
🚀  now let's let the dust settle slightly on develop and then I'll get this big ol' batch of commits tagged as stable soon.

PR #297, opened 2022-01-07
command to toggle open/closed the repl log window

philomates:
I recently switched over from iron.nvim and have been really enjoying conjure but I miss a quick way to toggle open and closed the REPL log window.

This PR introduces that logic to the `log.fnl` file and a new keymapping `<localleader> lg` to this new functionality.

Thanks so much for conjure btw! In the many years of using vim this is the first time I've been able to write any sort of non-trivial plugin code, and with a Fennel repl it was fun!


fnl/conjure/log.fnl
```diff
@@ -12,7 +12,8 @@
    require {sponsors conjure.sponsors}})
 
 (defonce- state
-  {:hud {:id nil
+  {:last-open-cmd nil
```
philomates, line 15: this could also have a default or be an entry in the config if folks think that makes sense

Olical: Yeah I was thinking that, I feel like vertical by default would be okay, it _feels_ like the most common use to me (although that won't be the same for everyone).

fnl/conjure/log.fnl
```diff
@@ -368,14 +370,25 @@
...
   (let [buf (upsert-buf)]
-    (close-hud)
-    (->> (nvim.tabpage_list_wins 0)
-         (a.filter
-           (fn [win]
-             (= buf (nvim.win_get_buf win))))
-         (a.run! #(nvim.win_close $1 true)))))
+    (a.filter (fn [win] (and (not= state.hud.id win)
+                             (= buf (nvim.win_get_buf win))))
+              (nvim.tabpage_list_wins 0))))
```
Olical, line 377: Ah so this is a refactor of `close-visible`, you've taken the finding bit and extracted it for reuse, nice.

fnl/conjure/log.fnl
```diff
@@ -368,14 +370,25 @@
...
+(defn close-visible []
+  (close-hud)
+  (close (find-windows)))
+
+(defn toggle []
+  (let [windows (find-windows)]
+    (if (a.empty? windows)
+      (when state.last-open-cmd
+        (create-win state.last-open-cmd))
+      (close-visible windows))))
```
Olical, line 391: I suppose if someone has many windows open they'll all close then the next toggle will reopen the last kind of log split. I think that's okay?

I this there's a problem here though:

 * Open the log in a tab
 * Go back to your code leaving the tab open
 * Toggle
 * The tab isn't "visible" so it's going to create a new window which will be yet another tab

So in the case where you have two tabs, your code and the log, and you're looking at your code, this mapping will just make more and more tabs 🤔

Olical:
Fixed conflicts with my own changes / other PRs and ensured it only tries to toggle when the last open command was vertical or horizontal split since a tab etc doesn't make sense for this right now. Although a tab toggle might be cool, that's basically ,lt then :q anyway.

We can always make this smarter in the future. Thank you for the addition!

PR #300, opened 2022-01-08
Add mapping to load the log buf into the current window

jsn:
I use this all the time. I used `nmap <localleader>lS <localleader>ls<C-W>o` for a while, but it's obviously not ideal (ugly, and won't work at all for someone actually using nvim windows).


Olical:
Nice this sounds useful. I _think_ I'll change the default mapping to `lo` for `log open` but this is good! Thank you!

jsn:
Cool, thank you! If 'll' is not available, I would prefer some mapping that ends in a key typed using left hand on 'qwerty', like 'lq' or 'le' or something. It's no big deal obviously, I can certainly always remap it locally :)

Olical:
I mean I kinda like `ll` as well, but I have two other mappings in PRs to merge right now and one is competing for `ll` space 😅  I need to take all three into account and find logical places for them all. But yeah I was assuming you'd remap this one straight back to `ll` if I did change the default 😄 I may not! I quite like it but I can see `log open (lo)` or even `log edit (le)` working well. I like mnemonic by default to help grouping and memorisation in new users.

Olical:
I'm going with `le` by default for `log edit` since that's the command that's essentially used. Obviously please do set it to `ll` if you prefer!

PR #302, opened 2022-01-14
Functionality to set log buffer cursor to top of last repl event

philomates:
Often when evaluating forms that have large output it would be useful to auto-jump to the top of that form so you can start reading the output from the start.

This PR implements that functionality and maps it to `<localleader>ll`. 

Comparable to leaving the auto-scrolling to the last line via typing `G` in the log window, you can enable the log window to always jump to the top of last eval event via `g:conjure#log#jump_to_latest#enabled=1`. Enabling this will disable auto-scrolling to the last line when the buffer is already at the last line, which might frustrate people.

A scrolling command hook is exposed such that when a jump to the top of the last event is performed the scroll command is run. Setting it to `zz` calls `"norm zz"` and centers the cursor in the window, `zt` sets it to the top, etc.

I still need to use this new functionality for a little bit to see what the best UX is, but wanted to share it in case others have thoughts or want to try it.


fnl/conjure/log.fnl
```diff
@@ -242,6 +242,15 @@
       (or buf (upsert-buf))
       (+ -2 (or extra-offset 0)) -1 false)))
 
+(defn jump-to-latest []
+  (let [buf (upsert-buf)
+        last-eval-start (nvim.buf_get_mark buf "E")]
```
philomates, line 247: I tried a local mark with `e` instead of `E` but it didn't seem to work for some reason. Not sure if there is a better way to do this

fnl/conjure/config.fnl
```diff
@@ -119,6 +120,8 @@
           :border :single
           :anchor :NE}
     :botright false
+    :jump_to_latest {:enabled false
+                     :scroll_command "zz"}
```
philomates, line 124: Also considered an enum of `:top | :center | :bottom` instead of a generic command here. Couldn't think of a good reason to do that though

fnl/conjure/log.fnl
Olical: Hmm not sure why _but_ I think maybe this is a good use for extmarks. These single character marks aren't suitable for automated use by plugins in my opinion, it's too likely to collide with a user thing or another plugin.

I _think_ extmarks are marks but for automated systems that won't collide with this narrow space of single character IDs.

doc/conjure.txt
```diff
@@ -625,6 +630,18 @@ Tip: Booleans in Vim Script are written as `v:true` and `v:false`.
             Marker to use to represent the end of a log fold.
             Default: `"}%~~~"`
 
+                                        *g:conjure#log#jump_to_latest#enabled*
+`g:conjure#log#jump_to_latest#enabled`
```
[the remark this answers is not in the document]
Olical: This specific line will only work if you're actually focused on the log window itself at the time of the eval. Were you in the log or another window at the time? And it's possible the custom autocmd stuff isn't working now, it's a little known area of Conjure not even I use.

fnl/conjure/config.fnl
Olical: The benefit of an enum would be that we abstract away the actual specific vim commands that are required and open up possibilities in the future. But using the raw command like this we allow people to get creative but we lose flexibility later if we need to do something a little smarter.

There's merits to both routes, I personally think the enum is the better option so the user could choose something that's hard for them to express themselves. So say `top` is really hard for some reason, like six normal mode buttons, we can hid that so each option is the same complexity.

Then if the log is reimplemented some day or has some other features we can upgrade how this backend is implemented while keeping the config the same abstract enum.

Olical:
I'm having a look at the enum and extmark work now

Olical:
Okay both suggestions implemented, let me know what you think! It's all based on extmarks now so there's no chance of it conflicting with one of the other marks a user might set.

PR #303, opened 2022-01-18
take into account line joining when setting eval result extmark

philomates:
after https://github.com/Olical/conjure/commit/6b441676a9c67380d6e43857f73d94b47d1291fb I'm seeing the following error in the log buffer when running tests via conjure
```
; run-ns-tests: integration.foo-test
; 
; Testing integration.foo-test
; localhost:33823 (/home/mates/.vim/plugged/conjure/lua/conjure/log.lua:367: line value outside range): .nrepl-port ....
; localhost:33823 (/home/mates/.vim/plugged/conjure/lua/conjure/log.lua:367: line value outside range): .nrepl-port ...
; localhost:33823 (/home/mates/.vim/plugged/conjure/lua/conjure/log.lua:367: line value outside range): .nrepl-port ...
```

This PR seems to fix it

### DIFF
--- a/fnl/conjure/log.fnl
+++ b/fnl/conjure/log.fnl
@@ -335,21 +335,19 @@
 
                     lines)
 
-            old-lines (nvim.buf_line_count buf)]
+            old-lines (nvim.buf_line_count buf)
+            new-line-index (if
+                             (buffer.empty? buf) 0
+
+                             ;; Replace one extra line if joining across fold markers.
+                             join-first? (if last-fold? -3 -2)
+
+                             -1)]
 
         (let [(ok? err)
               (pcall
                 (fn []
-                  (nvim.buf_set_lines
-                    buf
-                    (if
-                      (buffer.empty? buf) 0
-
-                      ;; Replace one extra line if joining across fold markers.
-                      join-first? (if last-fold? -3 -2)
-
-                      -1)
-                    -1 false lines)))]
+                  (nvim.buf_set_lines buf new-line-index -1 false lines)))]
           (when (not ok?)
             (error (.. "Conjure failed to append to log: " err "\n"
                        "Offending lines: " (a.pr-str lines)))))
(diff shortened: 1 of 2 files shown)


fnl/conjure/log.fnl
```diff
@@ -358,7 +356,7 @@
               jump-to-latest? (config.get-in [:log :jump_to_latest :enabled])]
           (nvim.buf_set_extmark
             buf state.jump-to-latest.ns
-            (a.inc old-lines) 0
+            (+ new-line-index old-lines 2) 0
```
philomates, line 359: the `2` is a "magic" number for me, without it the evaluation result doesn't show

philomates:
hmm, not sure actually if this is the right thing. I'm having trouble getting the feature working at all. I'll have to play with it a bit more

Olical:
Looking at this now!

philomates:
Thanks!

> Looking at this now!

Olical:
Let me know if the changes on develop fix this! I took my own approach instead of your PR just so I could understand the issue and hopefully fix the root cause while also guarding against future errors.

Worst case now should be that 1% of weird messages that produce line counts I don't expect won't jump to latest now instead of throwing errors.

Olical:
Oh I see! I've been trying to line it up with the first character of the actual output because if you think of this:

```fennel
;; Some thing!
{:foo :bar}
```

I'll _probably_ want to evaluate/edit/yank the `{:foo :bar}` or even jump to the matching brace with `%`. So I thought starting on the first character was usually the most useful thing. If moving to the very first character of the output is best then maybe we can do that instead? I just envision people "jumping to latest" then down a line every time, so I thought starting on that output line made more sense.

philomates:
https://github.com/Olical/conjure/commit/2fc39c5d80acdb14a150ed09e4a0da41bcd77dba seems to fix the print thing, thanks!

> I'll probably want to evaluate/edit/yank the {:foo :bar} or even jump to the matching brace with %

that makes sense. Honestly I didn't really think through the difference of std out stuff and eval results for this feature. Setting the cursor to the top of either makes sense for different reasons. Someone on slack suggested a way to jump forward/backward through eval results, so maybe something like that could help with this also. 

Let's leave it how it is and as I use conjure more in my dev work I'll try to think of a nice way to join all this

Olical:
Awesome! Closing for now and will get this released in a sec, we can always tweak more after using it for a while.

PR #308, opened 2022-02-06
Raw printing in test mode

qnkhuat:
Following https://github.com/Olical/conjure/pull/274.
Currently when set `let g:conjure#client#clojure#nrepl#test#raw_out = v:true` and run the tests, all output will be prefixed with a `;` this makes syntax highlight not working.

This PR removes that `;`.
Before
Screen Shot 2022-02-06 at 16 28 00

After
Screen Shot 2022-02-06 at 16 28 39


Olical:
Thought this through and I think it's the right move, thank you!

Olical:
Changed base to `develop`

PR #324, opened 2022-03-21
Check for each nREPL port file at each tree level (#317)

### LINKED ISSUE #317: Check for each nREPL port file at each tree level
As Hukka over on Clojurians slack pointed out, Conjure will pick the Clojure `.nrepl-port` file over the shadow-cljs one if a single `.nrepl-port` exists anywhere up your directory tree, so it's hunting for that first _then_ shadow-cljs if it can't find one. It should instead check each directory for each file in priority order.

The priority remains the same (Clojure then ClojureScript) but it should check for each in each directory as it travels, instead of travelling upwards twice.

stelcodes:
Fixes #317

Hi there Ollie! I'm trying to make my first Conjure contribution and I took a crack at this issue. This isn't ready for merging at all but I wanted to show you that I'm working on it. ~~I still need to get my nvim fennel setup functional and I haven't run any tests yet.~~

~~The change I made is basically so that `resolve-above` now takes a list of filenames and also has a 2-arity version that accepts a path. This path is expected to be a directory without a trailing slash which I believe is compatible with the current code. Then it goes up each level of the path hierarchy and searches for the nrepl files with the `findfiles` function already in place.~~

~~The surprisingly tricky bit is getting the parent directory from a path string since we can't use `vim.fn.findfile`'s upward searching functionality anymore. We could upwardly search for each nREPL file and then compare the depth of each match but that is probably less efficient. I added the `parent-dir` function which uses Lua's pattern matching to get the parent directory from a string. I couldn't find any neovim builtin that could do this and that was pretty surprising. Maybe there's a better way? I'm no (neo)vim API expert.~~

~~The `parent-dir` function takes a filepath separator string which should be dependent on the host OS so there needs to be a check for Windows but I'm not sure how.~~

Feel free to suggest something totally different than what I have here! Happy to change it all up.

### DIFF
--- a/fnl/conjure/fs.fnl
+++ b/fnl/conjure/fs.fnl
@@ -25,14 +25,44 @@
     (when (not (a.empty? res))
       res)))
 
-(defn resolve-above [name]
+(defn parent-dir [path]
+  (let [res (-> path
+                (split-path)
+                (a.butlast)
+                (join-path))]
+    (if (= "" res)
+      nil
+      ;; TODO: Needs to be Windows compatible?
+      (.. "/" res))))
+
+(defn upwards-file-search [orig-names orig-dir]
+  "Given a list of relative filenames and an absolute path to a directory,
+  return the absolute path of the first file that matches a relative path,
+  starting at the directory and then upwards towards the root directory. If no
+  match is found, return nil."
+  (var names orig-names)
+  (var dir orig-dir)
+  (var file nil)
+  (while (and dir (not file))
+    (let [name (a.first names)]
+      (if name
+        (let [res (findfile name dir)]
+          (if res
+            (set file res)
+            (set names (a.rest names))))
+        (do
+          (set names orig-names)
+          (set dir (parent-dir dir))))))
+  file)
+
+(defn resolve-above [names]
   "Resolve a pathless file name to an absolute path by looking in the
   containing and parent directories of the current file, current working
-  direcotry and finally $XDG_CONFIG_HOME/conjure"
+  directory and finally $XDG_CONFIG_HOME/conjure"
   (or
-    (findfile name ".;")
-    (findfile name (.. (nvim.fn.getcwd) ";"))
-    (findfile name (.. (config-dir) ";"))))
+    (upwards-file-search names (nvim.fn.expand "%:p:h"))
+    (upwards-file-search names (nvim.fn.getcwd))
+    (upwards-file-search names (config-dir))))
 
 (defn file-readable? [path]
   (= 1 (nvim.fn.filereadable path)))
--- a/fnl/conjure/client/clojure/nrepl/action.fnl
+++ b/fnl/conjure/client/clojure/nrepl/action.fnl
@@ -65,7 +65,7 @@
 (defn connect-port-file [opts]
   (let [resolved
         (-?>> (cfg [:connection :port_files])
-              (a.map fs.resolve-above)
+              (fs.resolve-above)
               (a.some
                 (fn [path]
                   (let [port (a.slurp path)]


stelcodes:
Oh wow I thought Fennel had much of Clojure's stdlib, I'm actually reading the Fennel docs now and I'm way off 😆

stelcodes:
Ok I actually took the time learn how Fennel and Aniseed work lol. I totally rewrote the PR and made sure it actually compiles. I haven't tested the plugin yet, but I think it's getting close.

fnl/conjure/fs.fnl
```diff
@@ -25,14 +25,44 @@
...
-(defn resolve-above [name]
+(defn parent-dir [path]
+  (let [res (-> path
+                (split-path)
+                (a.butlast)
+                (join-path))]
+    (if (= "" res)
+      nil
+      ;; TODO: Needs to be Windows compatible?
+      (.. "/" res))))
```
Olical, line 36: I think I have a OS path sep value somewhere in Conjure or Aniseed 🤔 can't remember where... I'll have a dig if you can't find it! You're right that this'll probably explode on Windows otherwise.

fnl/conjure/fs.fnl
```diff
@@ -25,14 +25,44 @@
...
+    (let [name (a.first names)]
+      (if name
+        (let [res (findfile name dir)]
+          (if res
+            (set file res)
+            (set names (a.rest names))))
+        (do
+          (set names orig-names)
+          (set dir (parent-dir dir))))))
+  file)
```
Olical, line 56: Might be neater to implement this through recursive tail calls since we can then bail early. Maybe not, but might be worth trying just to see if it reads really nicely since it doesn't involve setting mutable things. I'm happy to have a play with that refactor too, it's really just me playing lateral functional code golf.

I like the idea though! I'll have a play with it soon! Thank you for the effort!

stelcodes: I was thinking the same thing! That's how I had it in my first draft (it didn't compile because I was using a bunch of Clojure stuff like `loop`) but then I read in the Fennel docs that a Lua while loop has perf advanatages over Lua recursion (https://fennel-lang.org/reference#while-good-old-while-loop). Idk how much of a factor that is here. I guess you could technically be like 10000 directories deep in your filesystem but I don't know if that's such a common use case haha

fnl/conjure/fs.fnl
stelcodes: Yeah there's `aniseed.fs.path-sep` but I was wondering if it's more complicated for Windows since there's the `C:\\` or whatever at the beginning of the path. I don't have much experience with Windows (I actively try to avoid having to use it 😆) and I don't have a testing machine.

Olical: Instead of relative to the root we could always make this relative to the CWD... I think nvim has some functions to smush two paths together and it'll resolve all of the directory weirdness for us 🤔

fnl/conjure/fs.fnl
Olical: I'm not really worried about perf here tbh, it already looks pretty simple (which usually means fast!) and it's only run once every now and then. Normally it'll be checking two files about 5-6 levels deep (I would guess, on average). That's less IO that loading Aniseed's Lua files, a fraction of what my dotfiles do.

I think it'll be fine if we ran this every few seconds, but once a session? Totally fine, write it in a way that's easy to read and maintain, speed will come if we need it (although I already think this will be really fast! Most importantly, it'll be fast enough for the use case.

fnl/conjure/fs.fnl
```diff
@@ -25,14 +25,44 @@
...
   containing and parent directories of the current file, current working
-  direcotry and finally $XDG_CONFIG_HOME/conjure"
+  directory and finally $XDG_CONFIG_HOME/conjure"
   (or
-    (findfile name ".;")
-    (findfile name (.. (nvim.fn.getcwd) ";"))
-    (findfile name (.. (config-dir) ";"))))
+    (upwards-file-search names (nvim.fn.expand "%:p:h"))
+    (upwards-file-search names (nvim.fn.getcwd))
+    (upwards-file-search names (config-dir))))
```
Olical, line 65: Ah I like this explicit "search from file, then cwd, then config". I can't remember why I have the `config-dir` one in there, but there's probably a good reason. I think the original code was actually wrong, that `.;` string probably doesn't do what I thought it would...

So this is better in a few ways!

Olical:
Merged with some fixes, nothing too major. Also seeing some weird Fennel eval behaviour where it complains about too many locals... which means the REPL I'm evaluating into is getting too long. Which I fixed a LONG time ago 🤔 very odd!

PR #327, opened 2022-03-25
Change documentation command for fennel client

### LINKED ISSUE #326: Fennel (aniseed) documentation
Since 1.0.0 release `doc` special form was removed from fennel so `K` binding is useless.

It now uses `,doc` repl command to show documentation.

monkoose:
Should fix #326


Olical:
Great catch!

PR #335, opened 2022-04-18
Fix Common Lisp Getting Started URL

eHammarstrom:
* Added missing URL title description

Signed-off-by: Emil Hammarström <emil.a.hammarstrom@gmail.com>


Olical:
Ah good catch, thanks!

Issue #184, opened 2021-03-02
Julia and tree-sitter support


savq:
Hi, I found about Conjure a while ago, and while I can't say I've used it a lot, I find the premise great. This is by far the best REPL-plugin I've seen for Neovim.

I saw in the repo that non-lisp languages would be supported once Neovim has tree-sitter support, which is great! I've also been meaning to do some stuff with tree-sitter.

With all of that said, I'd be very interested in adding Julia support to Conjure.

I've already done a very rough draft [here](https://github.com/savq/conjure-julia). So far it only prints the text representation of the nodes to the nvim prompt, but these would be the same strings that Conjure would have to send to the REPL.

Since the folks over at [nvim-treesitter](link) have already done most of the hard work, I've just used some of the helper functions provided there. This also means that it'd have to remain a separate repo, so as to not add a dependency (and 0.5 requirements) here.

I wanted to know some opinions on this before moving forward. I know Conjure is waiting for 0.5 to stabilize, but I have time to spare these days, and I'd like to get something working in the mean time, at least for myself.

Olical:
Oh this is a wonderful example! I've been meaning to begin experimenting but I just don't have the time / brain space (especially recently), so your research here really lowers the barrier to entry. Thank you!

I could translate your examples into Fennel and integrate them into the `conjure.extract` module under a feature flag for people to opt into. Eventually that feature flag will be the default, then, some day, I'll delete the paren hunting code that's currently there.

Once that feature is in you'd be able to write a regular Conjure client for any language with the caveat that it asserts and throws an error if the tree sitter mode isn't enabled (or maybe your client module just exposes a `tree-sitter? true` value, probably that).

Olical:
`conjure.tree-sitter` is now a thing :smile: it's not hooked up to anything yet, but I've translated it to Fennel and added a link to your original repo. I hope that's okay? I've been putting thought into how we determine if we have the following requirements to use it:

 * Neovim 0.5 with tree-sitter
 * https://github.com/nvim-treesitter/nvim-treesitter (will this be part of the core nvim some day?)
 * The appropriate `:TSInstall [filetype]` already set up. Maybe I could call this for the user but that feels... presumptuous. Also it requires a bunch of things to work, so I'd rather not have Conjure checking if you have a working `gcc` copy :stuck_out_tongue:

[...]

PR #346, opened 2022-04-25
Add Julia client

TerseTears:
EDIT: I did some testing and most things seem to be working now. Let me know if I should squash the commits, etc.

This uses tree-sitter and the stdio approach. There are important drawbacks though that need fixing.

## What works:
- Evaluating current form
- Evaluating root form
- Evaluating file
- Inline results
- Showing documentation in log buffer

## Stdio issues :heavy_check_mark: 

EDIT 2: Passed the code to the `display` function and made some modifications to how results are displayed in the conjure and now the issue with conjure freezing is fixed.

EDIT: More specifically, the issue seems to be that `stdout::read_start` waits for output after each send request, but there isn't any. The issue basically boils down to the fact that julia prints the `nothing` value as "". And I took a look at the source, and, that decision is hard-coded...

- It seems that Julia doesn't send the prompt pattern (i.e. `julia>`) to stdout, hence the prompt needs to be set to the empty character `""`. Then the issue is that if a line prints nothing (such as `using Statistics`), conjure gets stuck.

## Possible Conjure issues :heavy_check_mark: 
EDIT: The issue was with the incorrect implementation of the `on-stray-output` part of the `start` function (commit daf32c4bf816d3d2ae24b5d7330f3be08b42cc0d).
- Upon the first send request, the log isn't properly updated and appears to get trimmed:
Screenshot_20220425_102133

## Other issues (Not greatly important)
- Tree-sitter: When there is an error in syntax, usually the root node is wrongly identified. For instance, when trying to send a line from the documentation:
Screenshot_20220425_103319
In the above, we do not want `julia>` to be sent as well.

- Conjure: There is a conflict with the [julia-vim](https://github.com/JuliaEditorSupport/julia-vim) plugin for the key `K`. julia-vim uses wordprg while `K` in conjure simply sends the doc to the log. Maybe a different mapping can be used by conjure?

## Ideas:
- There is a [RemoteREPL](https://github.com/c42f/RemoteREPL.jl) julia package which may provide the facilities to take a more robust approach than stdio.  [REPL](https://github.com/JuliaLang/julia/blob/master/stdlib/REPL/src/REPL.jl) itself is a module in Julia's stdlib which might be sufficient.
- Should new clients generally be separate plugins or is it okay to just merge all of them in conjure?

Related issue Olical/Conjure#184 and ping to the author @savq.


savq:
Hey @TerseTears,

This is really cool. Thanks for working on this and for letting me know.

> For instance, when trying to send a line from the documentation:
In the above, we do not want julia> to be sent as well.

The repl strips the prompt when pasting things, I guess you could do something similar like `(gsub str "^julia> " "")`, but that doesn't seem super important to me.

> There is a [RemoteREPL](https://github.com/c42f/RemoteREPL.jl) julia package which may provide the facilities to take a more robust approach than stdio

RemoteREPL makes a  julia-julia connection. At some point I tried using [Neovim.jl](https://github.com/bfredl/Neovim.jl) for a julia-nvim connection, which worked mostly fine. There's a small bug, but it should be fixable (the api has to be updated to 0.7 anyways).

> Should new clients generally be separate plugins or is it okay to just merge all of them in conjure?

I'd support making this a separate plugin for now.

Olical:
I'll look into this more today but quickly:

 * I'm happy with this being a PR or a plugin you'd install alongside. The only downside to the PR approach for you is that I'd like to review things before they go in, so a plugin might help you iterate faster but I do like all of the core languages being part of one big system that lazily loads what it needs. So do what works best for you but I like what you're doing right now 😄 
 * The on-stray-output receiving a string instead of a table of strings sounds familiar, another Neovim function had a breaking change recently I think, or made me use another function that works similar but returns a single string instead of a table of them. So I wonder if maybe adapting to that Neovim breaking change / fn deprecation caused a ripple of bugs like this.

Olical:
No commit squashing required! I like commits as they are as long as they're descriptive enough 😄 

Before I really dig into this it'd be great is someone else with a vested interest in this client could also confirm that your fork works for them, just to make sure there's enough information here for someone who knows the language and tools to work with it. If so I'll feel confident spending time verifying the support for myself / installing the toolchain / learning the basics of the language.

Olical:
@savq if you do have a little time, I'd love it if you could give us a +1 if you can get this working in your setup 😄

fnl/conjure/client/julia/stdio.fnl
```diff
@@ -0,0 +1,146 @@
...
+  ; remove last "\n" in s to connect display and s properly.
+  ( -> (string.gsub s "\n$" "")
+       (.. "|> display" "\n")))
+
+(defn unbatch [msgs]
+  (->> msgs
+       (a.map #(or (a.get $1 :out) (a.get $1 :err)))
+       (str.join "")))
+
+(defn format-msg [msg]
```
savq, line 52: Evaluating `println("foo")` results in
```
foonothing
```
So the result value needs a newline when there's something in stdout, but no newline otherwise.

savq (changes requested):
It's working very well for me. I'm running
```
julia 1.7.2
neovim 0.7.0
```

I wrote separate comments for the things that might need revision.

TerseTears:
I still have an issue with stray output. I don't know why the first time a form is sent to the REPL it goes to stray output. Might be an issue with the Julia REPL, but the output is received in disparate parts and since `log.append` has its own way of connecting these parts, the result isn't the same as the second time the form is sent to the REPL. I'm using the `join-first` option as an interim remedy. This is how the debug log looks like:

```fennel
# --------------------------------------------------------------------------------
# eval (selection): a = 2; b = 9; mat2 = [9 7] mat1...
# debug: send
"a = 2; 
b = 9;
mat2 = [9 7]
mat1 = [2, 4]

if(isnothing(ans)) display(nothing) end
"
# debug: receive
"out" nil "2"
# debug: msg in format-msg:
"2"
2
# debug: receive
"out" nil "
9
"
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"
9
"9
# debug: receive
"out" nil "1×2 "
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"1×2 "1×2 
# debug: receive
"out" nil "Matrix{"
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"Matrix{"Matrix{
# debug: receive
"out" nil "Int64}:"
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"Int64}:"Int64}:
# debug: receive
"out" nil "
"
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"
"
# debug: receive
"out" nil " "
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
" " 
# debug: receive
"out" nil "9  "
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"9  "9  
# debug: receive
"out" nil "7
"
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"7
"7
# debug: receive
"out" nil "2-element "
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"2-element "2-element 
# debug: receive
"out" nil "Vector{Int64}:
 2
 4
nothing"
# debug: on-stray-output-fn:
""
# debug: msg in format-msg:
"Vector{Int64}:
 2
 4
nothing"Vector{Int64}:
 2
 4
```

This isn't a great issue though, it's just sort of inconvinient to have to resend the form to see the results as displayed by the Julia REPL.

savq (approved):
Looks good to me!

The stray output thingie seems to only affect collections. Using `show` instead of `display` might fix it, but then the output wouldn't look like it does in the repl. I agree that it's not a serious issue.

Olical:
Okay, so, what do you think about merging and releasing this to get it in users hands so we can iterate? Can always ship it with the caveat of "it'll try it's best over stdio but it's hard to catch all output with this transport" then maybe one day this can be slowly replaced with some nREPL client or something for Julia if it ever gets one?

Happy with the code and behaviour right now? I'm going to defer to the judgement of @TerseTears and @savq since it's a new client and I don't know much about it. I'll handle any extra documentation / wiki / readme changes if they're required.

Olical:
Oh you might want to check out the use of `form-node?` around the existing codebase now as well as in `:h conjure`. You can use that as a hook in your client to get Conjure to skip non-evaluateable leafs in your code. So if you add `(defn form-node? ...)` into your client Conjure will call it every time the user tries to eval a form. If you return `false` for the given ts node argument it'll walk up one level and ask again.

You can turn the ts node into a string and try to work with that (like I do for all lisps, I check if they being and end with pairs of things) or you can walk around the node trying to find something that tells you it can be evaluated. I hope this helps improve the eval semantics! Maybe you weren't having issues with this anyway, but thought I'd mention it, it's just a way for every individual client to configure which kinds of ts nodes are selected as the "form under the cursor" for eval.

TerseTears:
> Okay, so, what do you think about merging and releasing this to get it in users hands so we can iterate? Can always ship it with the caveat of "it'll try it's best over stdio but it's hard to catch all output with this transport" then maybe one day this can be slowly replaced with some nREPL client or something for Julia if it ever gets one?
> 
> Happy with the code and behaviour right now? I'm going to defer to the judgement of @TerseTears and @savq since it's a new client and I don't know much about it. I'll handle any extra documentation / wiki / readme changes if they're required.

I think it's a matter of quantity vs quality in this case. The tree-sitter implementation allows writing clients essentially for all languages supported by tree-sitter. I've also been experimenting with a client for R. Implementing a client by the stdio approach is very easy, and I had little trouble figuring out what was needed. But it is always inferior to directly communicating with the REPL. Like, I couldn't figure out in the end how to switch the REPL mode by the stdio approach for Julia. Or in the case of R, a plugin that directly communicates with R such as Nvim-R (through a TCP server) can do a lot of other things as well.

I'm somewhat convinced the stray output issue is due to the Julia REPL itself, as hy and the R client do not exhibit the same behaviour.

My only remaining issue is that the inline result shows only the last line of the result but that is a separate issue with conjure I believe.

I also regret not having prefixed my commits with `julia: ` so if there's a least destructive way to fix that, please do let me know.

TerseTears:
> Oh you might want to check out the use of `form-node?` around the existing codebase now as well as in `:h conjure`. You can use that as a hook in your client to get Conjure to skip non-evaluateable leafs in your code. So if you add `(defn form-node? ...)` into your client Conjure will call it every time the user tries to eval a form. If you return `false` for the given ts node argument it'll walk up one level and ask again.
> 
> You can turn the ts node into a string and try to work with that (like I do for all lisps, I check if they being and end with pairs of things) or you can walk around the node trying to find something that tells you it can be evaluated. I hope this helps improve the eval semantics! Maybe you weren't having issues with this anyway, but thought I'd mention it, it's just a way for every individual client to configure which kinds of ts nodes are selected as the "form under the cursor" for eval.

This is certainly useful. Then I should see how to use the function properly and rebase unto develop.

Olical:
Let's merge this and get it into users hands + make it easier for anyone (including myself) to iterate on it 😄

Olical:
It works really well with tree sitter! Great job! :D I'm just adding it to all of the docs and wiki pages now.

PR #348, opened 2022-04-27
[WIP] Refactor and improve Common Lisp client

Dotrar:
This PR is to improve the Common Lisp client, mainly by refactoring the code and making it easier to read, and changing to the `slynk` client that is a more modern and maintained version of `slime` for common lisp.

# TODO:
- [ ] Confirm Slynk message encoding/decoding in `remote.transport.slynk`
- [x] Refactor out utility functions into `common-lisp.utils`
- [ ] Connect Common Lisp Client to slynk transport
- [ ] Have helper / install script in wiki page _ie: a one-liner to quickstart_
- [ ] Remove `swank`-y code.
- [ ] General tidy up before merging.

### DIFF
--- a/fnl/conjure/client/common-lisp/swank.fnl
+++ b/fnl/conjure/client/common-lisp/swank.fnl
@@ -3,6 +3,7 @@
              nvim conjure.aniseed.nvim
              bridge conjure.bridge
              mapping conjure.mapping
+             utils conjure.client.common-lisp.utils
              text conjure.text
              log conjure.log
              str conjure.aniseed.string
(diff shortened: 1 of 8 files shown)


Olical:
So what state would you say this is in right now? Still an improvement on what is there right now even if there's more you'd like to do? If you're kinda happy with how it behaves I'll be happy to merge and get it in the hands of users. Future changes are still of course welcome! 😄

Olical:
Oh actually this would be a breaking change right? Because it requires a different transport so the user would have to start their REPL differently?

Dotrar:
closing this PR because it's messy and out of date.

changing to slynk was a lot easier than i was expecting, check out my new PR here:
#400

Issue #365, opened 2022-05-30
Overhaul stdio based clients


Olical:
It's obvious that the various clients that aren't Clojure over nREPL are far more fiddly and brittle. Especially those that try to work over stdio.

The solution to this, I think, is to write a new stdio transport from scratch that's FAR more flexible. It should allow users to start and stop processes easily as well as provide MANY hooks into how it parses and understands the output.

The idea would be to remove the idea of "all we need is a regular express :)" and instead face the reality: Each REPL is different, we do _not_ have enough information since it's intended for humans, we NEED to use heuristics and hints that are unique to each REPL.

So a hypothetical future stdio client using this system (because it'll be reusable!) will be able to configure it with various functions that get given streams of lines (maybe one at a time) and they can respond with state and context to try to understand the output. We can use timeouts to confirm that the REPL prompt is back and the evaluation is complete.

Basically rather than assuming we can understand the output with a regex we should allow each client to parse the lines and make sense of them as they see fit with functions.

This issue is a placeholder so I don't forget to do this... but I'd like to make stdio clients feel as close to first class as I can. Even if it will never be as good as a proper nREPL server, I'd like to make the foundation solid for building more interesting interactions.

russtoku:
See [conjure.remote.stdio2: Rebuilding the foundation of stdio support](https://github.com/Olical/conjure/discussions/500) for what things might look like.

PR #369, opened 2022-06-20
Guile: Ensure that module is set before evaluation

taw10:
Using the Guile socket client, I noticed that my evaluations were not always getting evaluated in the correct context (module).  It was using the `BufEnter` autocmd, switching the module using the Guile REPL's `,m` command when the buffer changed.  It seems that this is not reliable, and sometimes it would fail to find the correct context and switch to the default module `(guile-user)` instead.  I think it might be somehow related to the presence of the log/HUD buffer, but I'm not sure at all.

This commit strips out all of the `BufEnter` stuff, and instead makes a separate `,m` command immediately before every evaluation.  The module name from `opts.context` in `eval-str`.

Whole file evaluations (`eval-file`) are run in the default module (`(guile-user)`), but if they begin with `(define-module ...)` then the end result is still correct: definitions within the file end up in the correct module.

The Racket client uses the same `BufEnter` system, and therefore almost certainly suffers from the same problem.

Extra discussion:
I tried to use the Guile REPL `,in` command (e.g. `,in (my module) (stuff to evaluate)`), but unfortunately this command requires the expression to be all on one line.  In the future, I'd like to have the Guile client inject a small "middleware" which can more elegantly handle evaluation of an expression within a specific module, and solve various other problems at the same time.  For example, the default REPL remembers all the evaluation results (so that they can be recalled with `$1`, `$2` etc), but this makes it possible to run out of memory fairly easily.  Definite link to #365 here.

### DIFF
--- a/fnl/conjure/client/guile/socket.fnl
+++ b/fnl/conjure/client/guile/socket.fnl
@@ -61,10 +61,15 @@
     (when (not (str.blank? clean))
       clean)))
 
+(defn- ctx-or-default [str]
+  (if str
+    str
+    "(guile-user)"))
+
 (defn eval-str [opts]
   (with-repl-or-warn
     (fn [repl]
-      (-?> opts.code
+      (-?> (.. ",m " (ctx-or-default opts.context) "\n" opts.code)
            (clean-input-code)
            (repl.send
              (fn [msgs]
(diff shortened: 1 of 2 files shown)


Olical:
This looks good! I introduced conflicts but I'll handle them.

PR #388, opened 2022-08-06
Add persistent environment support to lua neovim client

glyh:
As title. The current approach let user specify which is the desired behavior. By default it use debug tools to store locals across evaluation. This can be improved with other approaches. I think multiple approaches are preferred because each one has its own drawback.

This PR is a continuation of #292.

EDIT:

<del>I also add support for input prompt, but seems like there's some problems here.</del>
The problem in input prompt are being fixed by Olical in the develop branch.


glyh:
I'm not sure how you achieved this via coroutine because I didn't make a demo code that works... But it would be great if we can add that too.

glyh:
I just refactor the code a bit and add support to input prompt, but it seems problematic. And I guess this has something to do with repeat.vim because whenever I tried `io.read()` I got this:
```
':silent! call repeat#set(",e", 1)\n'
```

glyh:
I guess this is an issue. Directly call `(extract.prompt "Input required: ")` gives me the same result.

glyh:
@Olical I'm pushing it but did you see this PR?

Olical:
So the input prompt issue is something I'll solve with some unrelated work soon. Don't worry about that, it's a weird issue caused by hacks I had to do before Neovim had the proper Lua APIs to create mappings into Lua functions. Once I update how I perform a few of the mappings that issue will go away.

I've made small tweak to the output to make it valid Lua in the log buffer, another step on from your "return" lines you were inserting. I see why you had to do it, I'm prefixing results with `res =` for the same result but no syntax errors.

Seems better than what I had before! A little worried that overriding the io module will break a bunch of filesystem integration but we'll see. Still seems like an improvement! Merging!

Olical:
Merged to `develop`, will release soon.

glyh:
Sorry for my bad indentations, do you have a lint tool or something to fix them automatically?

glyh:
Yeah, it would be good if we can somehow redirect it. But we have to write some codes at neovim's side. I made a issue for it.

PR #400, opened 2022-08-27
Common lisp client improvement: Slynk support

Dotrar:
Common lisp client should support the updated, maintained and community supported SLYNK system, rather than the undocumented and outdated SWANK system.

This change replaces swank with slynk in Conjure, as well as some code clean-ups and improvements.

### Sly Functions
As a result of these changes, we can now easily support more *slyfun* (Sly Functions) in a clean and simple manor. Currently only two commands are defined, but this should be easily expanded.

```clojure
(defn eval-and-grab-output [msg context cb]
   (remote-execute :eval-and-grab-output msg context cb))
(defn connection-info []
   (remote-execute :connection-info))
```

There's around 100 Slyfunctions that can be incorporated, which can be found by looking for `(defslyfun` [in the slynk codebase](https://github.com/joaotavora/sly/blob/master/slynk/slynk.lisp).

#### Documentation

Documentation and reference for sly: https://github.com/joaotavora/sly/blob/master/CONTRIBUTING.md#s-exp-primitives
and of course the code itself.


Olical:
The concept sounds great and I'm find deprecating the old client and renaming/replacing it with this newer one. Sounds like a good change to me.

I'll give the code a read ASAP, I'll be inclined to LGTM and merge but will make sure it works on my system at the very least first.

fnl/conjure/client/common-lisp/parser.fnl
```diff
@@ -0,0 +1,151 @@
+(module conjure.client.common-lisp.parser
+  {autoload {a conjure.aniseed.core
+             text conjure.text
+             str conjure.aniseed.string
+             log conjure.log}})
+
+(defn- display-stdout [msg]
+  "Simple display to log"
+  (when (and (not= nil msg) (not= "" msg))
+    (log.append (text.prefixed-lines msg "; "))))
```
Olical, line 10: Can we move this to `fnl/conjure/remote/transport/slynk.fnl` or `fnl/conjure/remote/transport/common-lisp.fnl` for example? Then we should move any functions that are not ESSENTIAL to parsing out of it and into something in this `fnl/conjure/client/common-lisp` directory (such as `display-stdout`).

If the parser is stand alone and *pure* we can then write some tests for it easily and reuse it anywhere that similar data needs to be parsed.

fnl/conjure/client/common-lisp/parser.fnl
```diff
@@ -0,0 +1,151 @@
...
+
+  ;; match against these
+  (local slash-byte (string.byte "\\"))
+  (local quote-byte (string.byte "\""))
+  (local paren-open (string.byte "("))
+  (local paren-close (string.byte ")"))
+  ;; whitespace
+  (local space-byte (string.byte " "))
+  (local tab-byte (string.byte "\t"))
+  (local newline-byte (string.byte "\n"))
```
Olical, line 61: I'd recommend a `let` wrapping all of this instead of many locals, personal preference maybe. I normally use a big let and then have a mutable state map inside there that contains anything `var` related. That way I know all mutable state is prefixed essentially with a warning.

fnl/conjure/client/common-lisp/parser.fnl
```diff
@@ -0,0 +1,151 @@
...
+        quote-byte (open-close-quote b)
+        ; parens!
+        paren-open (open-paren b)
+        paren-close (close-paren b)
+        ; whitespace should finish word
+        space-byte (maybe-finish-word b)
+        tab-byte (maybe-finish-word b)
+        newline-byte (maybe-finish-word b)
+        ; everything else is inserted
+        _          (add-to-word b)))
```
Olical, line 132: Neat use of match, I don't think I've used that yet!

fnl/conjure/client/common-lisp/parser.fnl
```diff
@@ -0,0 +1,151 @@
...
+(defn- display-error [rest]
+  (let [ [eval-id b err & rest] rest]
+    (display-stdout (str.join "\n" err))))
+
+(defn parse-result [received]
+  (let [[return & rest] 
+        (unpack (parse-string-to-nested-list received))]
+    (match return
+      ::return (get-return-value rest)
+      ::debug  (display-error rest)
```
Olical, line 149: These look like Clojure namespace keywords, but I know it's actually `":return"` etc, can we use plain strings instead of Fennel's "keyword-ish" syntax here to be clear?

Olical (commented):
On the whole, I like it! Just a few style and code location comments. Would love to have that parser code extracted, in a shared space, pure and tested like the other transports/parsers.

fnl/conjure/client/common-lisp/parser.fnl
Dotrar: done

fnl/conjure/client/common-lisp/parser.fnl
Dotrar: I'll take the `let` because that's kinda neat. But I don't like the idea of using a map for mutable state within the function. 

Might be a clojure/CL thing (not that I'm good or professional with either!) but for the var's - the fact that they're out of scope anywhere outside of the parser function should kinda give the vibe that these are related to the state of the function in its operation, which can be fairly understandable in a parsing function.

The closures themselves are what it's all about. I think the idea of passing around a mutable state map that just gets dragged into each function is kind of a needless clutter when we can say "this is what `stack` is, and in our `foobar` function we add  to it, in our `foobaz` function we remove from it"

I'm still super young in lisp learning and this is your project so feel free to push back, this (revised) version is probably just the neatest I can envision it and I don't know if we need to be cautious of state when state won't be leaving this function anyway (unless there's some lua/fennel internal thing that I'm entirely unaware of)

fnl/conjure/client/common-lisp/parser.fnl
Dotrar: This has been done, I'll have a go with writing tests shortly.

Olical:
I'll handle those conflicts, but are you happy for me to merge this in it's current state or do you think it needs some more work? If you're using it and happy I'm happy to get it into more user's hands 🙂

UnkwUsr:
@Dotrar
\*polite ping\*

trev-dev:
Hey just an FYI, I used this in anger tonight and to be honest, it disappointed me every bit as much as swank did.  `(in-package)` doesn't seem to work in Conjure for either repl protocol.  If that's endorsement enough, you could merge this.

Olical:
@trev-dev worth checking the existing CL support again, I think I improved the `in-package` support https://github.com/Olical/conjure/discussions/557

I'm going to close this PR _for now_ but happy for it to be re-opened if we think it's still relevant or if someone wants to take what's here and carry the torch across the line.

PR #407, opened 2022-09-04
Strip shebang when auto require clojure

frankitox:
Hi! This little commit makes auto require work for babashka files that start with a shebang


Olical:
Merged to `develop`, thank you!

PR #410, opened 2022-09-09
allow custom user environment variables

sigmaSd:
This allows passing custom env in the config via `:env` this is useful for some repls like deno which needs `NO_COLOR=1`


Olical:
Merged to develop, thank you! I've altered it a little so we merge opts.env into the hard coded table, not the other way around. That prevents mutating opts.env (since a.merge! mutates values, hence the !) and allows callers to override those defaults.

PR #411, opened 2022-09-09
Add descriptions to mappings for compatibility with which-key

### LINKED ISSUE #205: Vim-which-key support
I see that `which-key` support was breifly added and then removed [here](https://github.com/Olical/conjure/commit/ba156dffc169a5c6e4c5244538c185a1fb128e0a).

Are you still planning on supporting it at some point?

If not would you mind saying why? Just so it shows up in google :)

Grazfather:
Fixes #205

This adds descriptions to the mappings, which are passed to `nvim_buf_set_keymap`, which is automatically picked up by folke/which-key.nvim

I haven't written all descriptions yet, want to make sure this sort of PR would be accepted.
tmux


fnl/conjure/config.fnl
```diff
@@ -105,6 +105,35 @@
     :eval_motion "E"
     :def_word "gd"
     :doc_word ["K"]}
+   :desc
+   {:log_split "Open log in new horizontal split window"
```
Grazfather, line 109: Maybe we want something like "Open **conjure** log..."?

Olical:
Oh nice! I like the change although I don't think they need to go under config, so I'll probably tweak that part but the idea is great.

The buf mapping function is a little fiddly and behaves differently depending on how many args you give it (which is bad!) because I've grown it over time and tried to keep it backwards compatible. I'll have a close read to make sure it works in all situations.

Olical:
Still more to do, but merged to develop + rewrote portions, I'm writing a new buffer local mapping function that's far better. Will slowly roll everything into that then delete `buf` in favor of `buf2` then rename. Your commit is merged though! I'll integrate desc with all clients and their mappings gradually. Lots to change to integrate this fully, but I'll get there!

Grazfather:
6f1fb260857e27d473d124454d40bdc2242013dd for posterity

I can add the rest of the descriptions if you like.

I think config sort of makes sense since it allows the description to be updated as easily as we already can the mappings.

Either way, thanks for merging!

PR #413, opened 2022-09-11
Add descriptiors for other mappings

Grazfather:
I noticed that with `buf2` on develop that missing descriptions show nothing
SCR-20220911-lrw

That's because you're using `:callback`. With 'legacy' `buf` you see the mapping, which is a bit friendlier:
SCR-20220911-lsk

So I figured I'd push some descriptions for the ones I didn't do before.

### DIFF
--- a/fnl/conjure/mapping.fnl
+++ b/fnl/conjure/mapping.fnl
@@ -26,23 +26,23 @@
    :log_reset_hard "Hard reset log"
    :log_jump_to_latest "Jump to latest part of log"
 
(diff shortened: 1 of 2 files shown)


Olical:
Nice thanks, I'll also default it to the command it's mapped to or something like that.

Grazfather:
My pleasure. Thanks for the merge and for the tool!

Olical:
Ah thanks, I'm not done with this yet though, I'll get to the others soon
and before releasing. Thanks for the diligence!

Issue #419, opened 2022-09-15
Log buffer filetype doesn't get set


austinmlv:
Expected behavior: after opening any Conjure supported source file I should be able to open a log buffer and see Conjure's output highlighted like the file being eval'ed. I should also have all the Conjure mappings available in the log buffer.

Actual behavior: Conjure's log buffer isn't getting the filetype detected despite having the right file extension. I open up a Guile '.scm' file, successfully connect to the repl, `<localleader>ls`, and then I get a plain text log buffer despite having a filename like `conjure-log-1234.scm`. Running `:set ft?` results in an empty string. As such there are no Conjure mappings set for the log buffer, no highlights, etc. However, running evaluations in the source file works as expected. Only filetype related things seem to be affected.

This only seems to happen to me with newer builds of Neovim. I'm running Neovim's master branch. Neovim v0.7.0  doesn't exhibit this behavior.

After playing around with `upsert-hidden` in [buffer.fnl](https://github.com/Olical/conjure/blob/a8686aa6f8760bd3cd4f219a8a4101af037c9d9b/fnl/conjure/buffer.fnl#L16) I can remedy this issue by calling `nvim.fn.bufload` on the newly created buffer.

Is anyone else having this issue? It seems to only affect recent Neovim builds on my machine. For what it's worth, `:help bufadd()` says that it doesn't load the buffer, therefore wouldn't the filetype settings/hooks not get run for the log buffer? From what I can tell Neovim was running the filetype hooks in earlier builds after running `nvim.fn.bufadd` and now it's not.

If this is indeed an issue here's a potential fix:

Olical:
Yes, I think Neovim have introduced a bug or breaking change. Thank you for finding a potential fix, I'll dig into this soon, should get some OSS time this weekend. I've heard this from a few people now but only on unstable / master branch Neovim. I only really run the latest stable release so I haven't noticed this yet myself.

austinmlv:
Thanks for the reply Olical. I just submitted a potential [fix](https://github.com/Olical/conjure/pull/420#issue-1374965223).

Olical:
Merged, can someone confirm that this is okay now on the `develop` branch? Thanks!

PR #420, opened 2022-09-15
Explicitly load newly created log buffer

austinmlv:
Possible fix for #419


Olical:
Happy to merge this, did it fix it for you? I couldn't reproduce it on my nvim version.

sigmaSd:
This makes the lsp active on the log buffer as well which I think is undesirable.

PR #423, opened 2022-09-17
Moved comment-form? fn to tree-sitter.fnl

eccentric-j:
Per our discussion in discord earlier today, only the clojure client defined the comment-form? function. Which meant when using a root form eval on a form in the following:

```fennel
(comment

  ;; <localleader>eR  
  (print (vim.opt.filetype:get))
  
  )
```

It would eval the whole `(comment ...)` which always resulted in nil. This PR moves it into tree-sitter.fnl and updates the root-form defn to call comment-form? directly instead of the optional client dispatcher.

I don't anticipate many negative side-effects as it's doubtful many users are purposely trying to eval root comment forms, and it doesn't change their behavior.


fnl/conjure/tree-sitter.fnl
```diff
@@ -67,7 +71,7 @@
     (if
       (document? node) nil
       (document? parent-node) node
-      (client.optional-call :comment-form? parent-node) node
```
Olical, line 70: Ah, so I now realise there's a bit of a change to make here but I'll merge this and make the change myself.

I want to keep this line as it was but make it so that each client can reuse a common function to check for lisp style comments. But they can also reuse some C-like language comment checker function, for example. So we reuse the code, but each client gets to pick the function.

eccentric-j: I see, that makes sense to me.

eccentric-j: Should I resolve these merge conflicts before you merge?

eccentric-j: Just pushed a fix for the merge conflicts.

Olical:
Merged and modified! Thank you!

Issue #416, opened 2022-09-12
Conjure ignores `;` in Julia


usmcamp0811:
When using Julia, Conjure seems to ignore `;` placed at the end of a line. In Julia it is used to suppress the returning of the lines output. This is a problem because when you open a 15,000 line CSV into a DataFrame, Conjure kind of craps the bed and becomes really slow as it tries to print every single line. 

Example code to return the problem. 

```julia
using Downloads
using CSV
using DataFrames
example_csv = Downloads.download("https://raw.githubusercontent.com/JuliaAcademy/DataScience/main/programminglanguages_CSV.csv")
data = CSV.read(example_csv, DataFrame);
```

if you run the above in Conjure it will return the dataframe into the log. If you run it in a REPL it will not return anything (the desired result). 

This is a hack to not return something 

```julia
using Downloads
using CSV
using DataFrames
example_csv = Downloads.download("https://raw.githubusercontent.com/JuliaAcademy/DataScience/main/programminglanguages_CSV.csv")
# this will cause it to return the `example_csv` path instead of the `DataFrame` 
data = CSV.read(example_csv, DataFrame); example_csv
```

I have tested this with `:ConjureEval` in normal and visual modes. I've done `<localleader>eb` and get the same results. Pretty much I've tried it with every eval method I saw as being made public in Conjure.

usmcamp0811:
So as to add a little bit of info to this in the event I get hit by a bus or future people look back.. I spent a good bit of yesterday poking and prodding things to see what I could learn. The biggest take away I could gather is that interfacing with the Julia REPL via `stdin` will completely ignore all `;`. The thing that will squash the out put if ended with a `;` is a function built into the Julia REPL and defined around [here](https://github.com/JuliaLang/julia/blob/ebe3159642b59f482616ebbc82b612f7fd4892a2/stdlib/REPL/src/REPL.jl#L1361). When we pipe or otherwise send things to `julia` via `stdin` it would seem we are not hitting the REPL at all, but rather just going straight to the compiler, hence we still get back big long arrays even when we don't want or need them. 

All is not lost as mentioned elsewhere on here I looked into the [RemoteREPL](https://github.com/c42f/RemoteREPL.jl) and its legit! I think we can use this to give comparable support to Julia as is for Clojure. Somewhere I found this shell script and I made some modifications to it (below) and it can currently take in a string argument that will 100% be sent to the remote REPL and will be processed as expected. It will then `@show` the like where its evaluating things. 

[...]

Issue #380, opened 2022-07-15
don't log giant objects/format julia output the same way the repl formats things!


josephsdavid:
I am running into a slight issue at least with the Julia client, everything works great, but, if i run a bit of code that returns say a table with 600k rows, every single row is stored in the log, which causes a conjure to eat up all my cpu!

MWE:

```julia
import Pkg
Pkg.add("DataFrames")
using DataFrames

x = DataFrame(1:1e7)
```

When I run this in the julia repl, I get:

```
10000000×1 DataFrame
      Row │ x
          │ Float64
──────────┼────────────
        1 │  1.0
        2 │  2.0
        3 │  3.0
        4 │  4.0
        5 │  5.0
        6 │  6.0
        7 │  7.0
        8 │  8.0
        9 │  9.0
       10 │ 10.0
       11 │ 11.0
       12 │ 12.0
       13 │ 13.0
       14 │ 14.0
       15 │ 15.0
       16 │ 16.0
       17 │ 17.0
    ⋮     │     ⋮
  9999984 │  9.99998e6
  9999985 │  9.99998e6
  9999986 │  9.99999e6
  9999987 │  9.99999e6
  9999988 │  9.99999e6
  9999989 │  9.99999e6
  9999990 │  9.99999e6
  9999991 │  9.99999e6
  9999992 │  9.99999e6
  9999993 │  9.99999e6
  9999994 │  9.99999e6
  9999995 │  1.0e7
  9999996 │  1.0e7
  9999997 │  1.0e7
  9999998 │  1.0e7
  9999999 │  1.0e7
 10000000 │  1.0e7
   9999966 rows omitted
```

However if i evaluate this block with conjure, every item is logged!!

Olical:
Ah that sucks! Hmm I wonder if there's an option we can set to tell the REPL not to print everything? Or do we have to receive everything from the REPL and then do the truncating inside Conjure 🤔 

I'd like to set some sort of global that auto truncates, that's what I do with Clojure / CIDER and it does some clever nested data truncation which is far smarter than anything I could do inside Conjure.

josephsdavid:
> Ah that sucks! Hmm I wonder if there's an option we can set to tell the REPL not to print everything? Or do we have to receive everything from the REPL and then do the truncating inside Conjure 🤔 
> 
> 
> 
> I'd like to set some sort of global that auto truncates, that's what I do with Clojure / CIDER and it does some clever nested data truncation which is far smarter than anything I could do inside Conjure.

[...]

PR #435, opened 2022-11-02
Fix  Conjure ignores ; in Julia #416 

usmcamp0811:
Since the Julia REPL we are using doesn't care about `;` this is a "duct tape-y" solution that uses regex to match a `;` at the end of the line and replace it with `; nothing;` which will ultimately suppress the output of the given line. 

As stated in the discord channel this works if you eval the entire line manually. `<localleader>ee` will not send the `;` to the REPL for evaluation.

relates https://github.com/Olical/conjure/issues/416
relates https://github.com/Olical/conjure/issues/380


Olical:
So to clarify, if you use `,ee` it still won't pick up `;` even with this change, so you need to remember to eval the whole line if you want to have `;` work, even after this change.

Because this changes `;` from not working at all to working if you eval the line with a visual selection? If so I can tweak this a little further and try to get the tree sitter integration to include `;` in Julia. Happy to have a go at that if this solves the "sending a `;`" problem.

Olical:
I like it! Merged, will have a look at the treesitter stuff myself too. If it really won't EVER capture a `;` then yeah I agree with your idea, we should check the text for a trailing one.

PR #445, opened 2022-12-08
[debugger]: add log for completed init

awwalker:
In my anecdotal experience the time between starting the debugger and the initialization actually completing is long enough that attempting to add breakpoints and re-trigger a function often completes before the debugger has registered. This adds a log to the buffer for the completed initialization so you _know_ when the debugger will be active.

```
; Initialising CIDER debugger
; eval (root-form): (add 5 9)
14
; eval (current-form): (defn add [a b] #break (+ a b))
#'scratch.visa-settlement/add
; eval (root-form): (add 5 9)
; CIDER debugger initialized
...
```


Olical:
This is neat, thank you!

PR #451, opened 2022-12-30
IPython Notebook support

CamdenClark:
This PR is adding in IPython support, which would make it possible to run `ipynb` (IPython Notebooks) using conjure + jupytext.vim

Basically what you would have is a markdown document with markdown code blocks you can execute using conjure. When you save that, it gets converted back to the ipynb format.

Wanting to get early feedback!

Some TODOs:
- *Need help here--banging my head against the wall* I'm trying to get [bracketed paste](https://cirw.in/blog/bracketed-paste) working to be able to do multi-line evaluation. I really can't get Conjure to send the right escape code it seems :( I'm able to manually type this in at the IPython terminal and it works, so not sure what to do here. [Iron.nvim seems to do the same thing]((https://github.com/hkupty/iron.nvim/blob/f4131638454aeb96cbd92652dd8c396e48eb4118/lua/iron/fts/common.lua#L28)

Some ideas of what this could be: readline config being `/dev/null`, so bracketed paste is not supported.

Just really not sure how to really dig in here, so some debugging tips would be helpful.

- Need to get the output right to separate what gets logged vs what the output of your cell is.

- Lots of cleanup


thecontinium:
Not sure I can help but I'm going to be switching between Ipython and Clojure so this would be really good to see working.

CamdenClark:
> Not sure I can help but I'm going to be switching between Ipython and Clojure so this would be really good to see working.

I'll set aside some time this week to play around with it.

thecontinium:
Great - If you need someone to test it out just give me a shout.

Cheers

--
The Continium

On 9 January 2023 at 15:16:23, Camden Clark ***@***.******@***.***>) wrote:

Not sure I can help but I'm going to be switching between Ipython and Clojure so this would be really good to see working.

I'll set aside some time this week to play around with it.

thecontinium:
You might want to take a look at this [Femaco](https://github.com/AckslD/nvim-FeMaco.lua)

Using this out of the box with [Juptext](https://github.com/goerz/jupytext.vim) enables conjure evaluation of the python code blocks in a floating window; I think even configurable to it's own window. Once the code block is open the conjure log buffer can be opened and all the conjure evals etc are available.

The advantage of this is that you get all the lsp + diagnotics + code completion for python all setup for you for free by the file type `python` in the new window.

russtoku:
I tried [ca6af22](https://github.com/Olical/conjure/pull/451/commits/ca6af2292f060410e776c298b300fab972ee5ee8) but it didn't seem to do anything. Am I missing something?

russtoku:
I did some experimenting this past week and can confirm that the bracketed paste mode approach doesn't work with the stdio client. I can also confirm that [hkupiron.nvim](https://github.com/hkupty/iron.nvim) uses a neovim terminal to run a repl in and that's why bracketed paste mode works for it. Hope this helps.

CamdenClark:
> I did some experimenting this past week and can confirm that the bracketed paste mode approach doesn't work with the stdio client. I can also confirm that [hkupiron.nvim](https://github.com/hkupty/iron.nvim) uses a neovim terminal to run a repl in and that's why bracketed paste mode works for it. Hope this helps.

That's really frustrating but good to know -- sorry I haven't been active on this PR.

Olical:
Pushed changes to the stdio client today that Base64 encodes the code as it enters the REPL and decodes it as it evaluates it, so we never have to send any kind of binary or escape codes into the REPL anymore. That may be very useful here too, I hope I've fixed a bunch of the original Python eval experience with this, you can reuse the Lua Base64 functions I added as part of this too.

Olical:
Also I think we could make minor adjustments to the existing stdio wrapper code (despite it's other shortcomings) so it starts a nvim terminal instead of just a headless process. I wonder if there are similar interfaces for both, I just thought headless would be best since the user might break things if they type into the REPL directly, keeping it headless seemed safest but was maybe a mistake.

russtoku:
I've been reading Mickey Peterson's Mastering Emacs article, [Shell & Comint Secrets: History commands](https://www.masteringemacs.org/article/shell-comint-secrets-history-commands). Using a terminal is appropriate where you want to take advantage of how terminals works (rows and columns of characters). Since, @Olical fixed the issue with sending multiline Python code to a child process, I think there's less of a need to support interaction with a terminal. That may be my own short-sightedness.

CamdenClark:
Thanks for taking the lead on the base64 stuff, that's really awesome and unlocks getting this to something people can actually use.

I'm going to take this PR out of draft. There are still a few outstanding things:

1. Evaluate file is unimplemented
2. When you evaluate a cell with multiple expressions, it won't print the output of the last command. This is why bracketed paste would be so much more effective: it could work much more transparently like how the client works.
3. We never return the last expression because of how ipython formats stuff (maybe this is shortsighted and we should anyways? For example, if you have something like this get returned through stdout
```
Hello world!
array([[0, 0, 0, 0],
            [0,0,0,0]])
```
How do you know to return the last array object, even though it's on multiple lines?

So it's tricky, but this is still an incredible workflow for getting productive on ipython notebooks.

fnl/conjure/config.fnl
```diff
@@ -53,7 +53,7 @@
    :path_subs nil
    :client_on_load true
 
-   :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust]
+   :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust :ipynb :markdown]
```
CamdenClark, line 56: Is there a better way to do this? I only want this to work on files with the `ipynb` extension

russtoku: I think that only the :ipynb filetype would need to be added because when editing a Markdown file there would not be an associated REPL.

Would this IPython Notebook support require the `goerz/jupytext.vim` plugin along with jupytext?

Olical:
I need to have a proper read and review of this PR, but please see my recent Python changes around evaluating things with newlines inside them. I avoided bracketed paste which I also couldn't get working by Base64-ing the code before it heads into the REPL then decoding it on the other side before evaluation. It doesn't sound pretty, but it works great and avoids any readline / newline weirdness completely.

I still want to write a whole new stdio support layer from the ground up that makes all of this a lot better, but that's not started really yet. I started the file but haven't made any solid progress.

Issue #460, opened 2023-02-05
Expose command that uses extract/legacy-word and/or provide option to use it by default


Kamholtz:
tpope/vim-sexp-mappings-for-regular-people re-defines the word text-object to include punctuation such that `vim.fn.expand` is considered a single word. Prior to Conjure using treesitter to extract the word text object, it had abided to vim-sexp-mappings-for-regular-people's new definition of a word. Now, with treesitter enabled and vim-sexp-mappings-for-regular-people installed, executing `ConjureEvalWord` with the cursor on the "v" in `vim.fn.expand` would eval just eval `vim` instead of `vim.fn.expand`. I found it quite useful to be able to eval the entire word including the punctuation, particularly when configuring Neovim. 

Would you consider adding a command that uses `extract/legacy-word` and/or provide a per language option for defaulting to `extract/legacy-word`?

Kamholtz:
I have made a pull request (#461) to illustrate what I have in mind and am happy to make adjustments if necessary!

Olical:
I think what we _really_ want here is to adjust the "extract word" tree sitter implementation so it takes the whole word into account, like the sexp mappings plugin does. I'd rather not use the legacy word function if we can avoid it but will still keep an open mind. Avoiding another config option would be nice too, there's already so many!

I'll have a quick play with the TS code, I have a feeling it used to work like you described at some point, maybe the TS definition changed at some point 🤔 it's annoyed me personally too, definitely prefer the whole symbol idea, we're not dealing with words, we're dealing with _symbols_ which can contain `.`s etc.

Olical:
Should be fixed! We now look up one level at the parent to check if it's type is `multi_symbol`. If so, we use that node instead of the lowest child that the tree sitter API reported initially.

Kamholtz:
I agree that evaluating symbols by default is preferred! I am now using develop branch and it is working as expected. I will let you know if I notice anything odd. Thank you for looking into this!

PR #461, opened 2023-02-05
Ck/feat/2023/02/05/legacy word

Kamholtz:
A rough attempt at providing a config option for using legacy-word by default and also providing a command/key-binding that that uses legacy-word. 

#460


Olical:
Released my fix 🎉

Issue #437, opened 2022-11-16
Python SyntaxError: EOL while scanning string literal


pheeria:
Hi 🖖 

Thanks for the awesome plugin! And for the recent Python support, looks like I don't need `vim-repl` anymore 😄 
I've found the following issue while using Conjure with Python.

This code runs as expected:
```python
"hey" + "\n" + "ho"

### Output

# eval (root-form): "hey" + "\n" + "ho"
'hey\nho'
```

But putting this concatenation in a function fails:
```python
def func():
    return "hey" + "\n" + "ho"

### Output

# eval (root-form): def func(): return "hey" + "\n" + "ho"
# (out)   File "<stdin>", line 1, in <module>
# (out)   File "<string>", line 3
# (out)     return "hey" + "
# (out)                    ^
SyntaxError: EOL while scanning string literal
```

The same is also true for using `join`:
```python
"\n".join(["hey", "ho"])
```

All of the above code runs when executed from console.
From my very limited understanding `escape-strs` from `fnl/conjure/client/python/stdio.fnl` should not escape already escaped characters. Please let me know if I can help further.

```
nvim: 0.8.0
conjure: latest master, 8.10.22
```

Olical:
Thanks for the report! What's fun is the log is telling us Conjure selected all the right code with tree sitter:

```
# eval (root-form): def func(): return "hey" + "\n" + "ho"
```

The issue is _probably_ some escaping that's required when sending Python code across this stdin stream. Like the `\n` needs to become `\\n` I think.

Olical:
Current investigation has revealed that Python's `exec()` which we _need_ to rely on to be able to evaluate your code correctly doesn't like `\n` character's in it's source code. So `\n` in a normal Python source file is interpreted differently when sent into a REPL through the `exec()` call.

If I don't use `exec()` then your REPL will hang and throw errors if you try to eval loops / function definitions etc that require very specific sequences of tabs and newlines (and won't allow two new lines in a row!). So I think I either need to overhaul how the Python REPL works, use a completely different Python REPL or maybe find a way to pipe your code into a `/tmp` file and then load that in the REPL.

If I do that though you'll lose state between your evals, so I'd need to find a way to:

 * Write your code to eval into a tmp file.
 * Tell Python to load and eval that file.
 * Somehow extract the context of that eval into your current REPL env so the defined functions etc are shared between this eval and the next.

Maybe there's a Python function that lets me run a file of Python code and capture the scope for future reference 🤔 

[...]

PR #464, opened 2023-02-16
fix for issue #437

russtoku:
I think I've solved the problem with embedded newline characters in Python source code. All of the examples in `dev/python/sandbox.py` evaluate correctly. I haven't tried it with other Python files yet. Just wanted to get something working for others to try on their code.

### DIFF
--- a/fnl/conjure/client/python/stdio.fnl
+++ b/fnl/conjure/client/python/stdio.fnl
@@ -93,11 +93,17 @@
 
 (defn- escape-strs
   [s]
-  (string.gsub s "\"" "\\\""))
+  ;; Split s into multiple lines based on newlines.
+  ;; Escape internal escaped newlines in each line if any.
+  ;; Return lines joined backed together with newlines as a single string.
+  (let [s2 (str.split s "\n")
+        lns (icollect [i val (ipairs s2)]
+              (string.gsub val "\\n" "\\\\n"))]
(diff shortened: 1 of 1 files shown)


Olical:
So definitely an improvement! Although I've found another edge case 😬 I'm worried there will be many while we have to mess with new lines in any capacity.

```python
def newline_in_function_bug():
    return 'hey\n\n' + "\\n" + '\n' + 'ho'

newline_in_function_bug() # 'hey\n\n\nho'
```

The `\\n` should result in a `\n` being printed inside the string, but not an actual newline, a separate `\` and `n` character if I'm not mistaken? I'll have a look at your changes some more and see if I can add to it to catch this too.

Really I need to actually rewrite the stdio support from the ground up so we can deal with these things in a better way.

Olical:
The output from a plain Python REPL:

```python
>>> def newline_in_function_bug():
...     return 'hey\n\n' + "\\n" + '\n' + 'ho'
... 
>>> newline_in_function_bug()
'hey\n\n\\n\nho'
```

Olical:
https://media.giphy.com/media/TI0oUTrAKt3Ko/giphy.gif

Olical:
Cunning plan, executed. Base64 encode the source in Lua, decode it within the REPL, never deal with encoding issues again. 🚀 

We still have two eval styles depending on if you're evaluating an "expression" or not... maybe we can just delete those now and just have the one system?

russtoku:
Awesome! Many thanks!

PR #469, opened 2023-02-28
Python client and HUD obey global config settings

russtoku:
Sorry but I wasn't thinking and should have made separate pull requests for these small changes.

- The Python client  obeys the g:conjure#client_on_load configuration setting and doesn't automatically start up a REPL.
- The HUD is disabled when g:conjure#log#hud#enabled configuration setting is false.


russtoku:
Apologies, I just noticed that when I start up nvim and `g:conjure#log#hud#enabled` is set to false, the HUD behaves as it does when the setting is true.

However, if I toggle the setting to true then back to false, the HUD obeys the setting and doesn't pop up when the log buffer is opened in a split or hidden.

I'll need to look deeper.

russtoku:
Closing because I'm going back to the drawing board.

PR #470, opened 2023-03-01
Python client improvements

russtoku:
The Python client now obeys the `g:conjure#client_on_load` configuration setting. It won't start a REPL when the first Python source is loaded.

Also, <localleader>ee (evaluate current form) behaves in a manner more intuitive to Python fans. <localleader>ee on the first line of this code evalutes (5 + 7 + 9) not (5 + 7):
```
5 + \
7 + \
9
```


russtoku:
Apologies but I didn't realized that another fix to make the Python client behavior a bit more intuitive for me and hopefully other Python fans would piggy-back on this PR.

Olical:
Ohh this looks good! I'll give it a go locally but I think this should be safe. Thank you very much!

russtoku:
Thank you for creating Conjure and moving it along! I love it!!

Issue #477, opened 2023-04-04
vim.treesitter.require_language() is deprecated, use vim.treesitter.language.require_language() instead.


feoh:
Using neovim master, I get this error on startup now:

```
vim.treesitter.require_language() is deprecated, use vim.treesitter.language.require_language() instead. :help deprecated                                                                      
This feature will be removed in Nvim version 0.10     
```
Thanks!

Olical:
@rhcher fixed this in https://github.com/Olical/conjure/pull/479 🎉 it's on develop.

PR #479, opened 2023-04-09
fix(deprecation): update treesitter function

rhcher:
see https://github.com/neovim/neovim/pull/22761
vim.treesitter.query.get_node_text is deprecated, use new api instead

also resolve https://github.com/Olical/conjure/issues/477


Olical:
Oh thanks for doing this! Really helpful, I'll merge and may refactor slightly to make it obvious where the "supporting two Neovim versions" code is sitting. Losing track of that sort of thing is scary.

PR #487, opened 2023-05-05
Allow Scheme client to work with .ss files

### LINKED ISSUE #485: Support for Gerbil Scheme
I'm attempting to add support for Gerbil scheme (a wrapper around Gambit), but am seeing some really bizarre behavior from Conjure. I've created a new folder in `clients/` called `gerbil` with a `stdio.fnl` file in it largely copied from `scheme`. There's some peculiarities with the raw scheme backend, as it seems to look for `.scm` files, which aren't typically used for Gerbil which uses `.ss` files.

I'm testing with a simple Gerbil file:
```scheme
(export main)
(def (main . args)
    (displayln "Hello"))
```
and receive this error:
```
E5108: Error executing lua: Vim:Error executing Lua callback: Conjure client '' doesn't support function: eval-file
stack traceback:
...
```
I've elided the stacktrace, as it has nothing helpful in it. It seems Conjure is not detecting which backend it should use, it's very not clear how it even selects the backend. Assistance with this would be appreciated, otherwise the next best thing is some documentation.

chiefnoah:
The Scheme client expected a .scm file to bind to a buffer, however some Schemes such as Gerbil prefer the .ss extension.

Fixes #485


Olical:
Lovely! Thanks!

PR #498, opened 2023-05-25
Fix for issue #497 Scheme client

russtoku:
Sorry, had to create another branch in my fork to isolate my other PR from this one.

I copied the interrupt function from the Racket client instead of the Python client because the Racket one uses the `send-signal` function from the base functionality in`remote/stdio.fnl`.

PR #499, opened 2023-05-28
Use repl.send-signal to interrupt stdio REPLs

russtoku:
This is a minor nit but I noticed that the stdio-based clients use two equivalent implementations of the `interrupt` function. This PR makes the implementation the same for these clients:
- Hy
- Julia
- Python
- Racket

by using the:
- `send-signal` function defined in `remote/stdio.fnl`
- `vim.loop.constants.SIGINT` constant instead of the number 2
- `comment-prefix` of the client's language

Hopefully, this provides consistency and emphasizes the use of the base client functionality of `remote/stdio.fnl`.

I didn't include the Scheme client because there is https://github.com/Olical/conjure/pull/498 which adds the ability to interrupt the REPL.

### DIFF
--- a/fnl/conjure/client/hy/stdio.fnl
+++ b/fnl/conjure/client/hy/stdio.fnl
@@ -155,8 +155,9 @@
   (log.dbg "sending interrupt message" "")
   (with-repl-or-warn
     (fn [repl]
-      (let [uv vim.loop]
-        (uv.kill repl.pid uv.constants.SIGINT)))))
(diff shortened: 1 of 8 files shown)


Olical:
Okay, I _think_ that's all of your changes merged to develop fairly cleanly. Please let me know if something seems off! I'm going through myself too.

russtoku:
All seems in order. Thanks!

PR #510, opened 2023-06-21
Add installation instructions for lazy

Quitlox:
Small addition to the README for installing the plugin via lazy.nvim.

Though the simple setup is as easy as with any other plugin manager, getting conjure to lazy load took some time to figure out, so others might appreciate the instructions.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -96,36 +96,7 @@ To take advantage of lazy loading, a bit more involved configuration is required
 ----
 return {
     "Olical/conjure",
-    cmd = {
-        "ConjureEval",
-        "ConjureSchool",
-        "ConjureConnect",
-        "ConjureClientState",
-        --
-        "ConjureLogSplit",
-        "ConjureLogVSplit",
-        "ConjureLogTab",
-        "ConjureLogBuf",
-        "ConjureLogToggle",
-        "ConjureLogResetSoft",
-        "ConjureLogResetHard",
-        "ConjureLogJumpToLatest",
-        "ConjureLogCloseVisible",
-        --
-        "ConjureEvalCurrentForm",
-        "ConjureEvalCommentCurrentForm",
-        "ConjureEvalRootForm",
-        "ConjureEvalCommentRootForm",
-        "ConjureEvalWord",
-        "ConjureEvalCommentWord",
-        "ConjureEvalReplaceForm",
-        "ConjureEvalMarkedForm",
-        "ConjureEvalCommentForm",
-        "ConjureEvalFile",
-        "ConjureEvalBuf",
-        "ConjureEvalMotion",
-        "ConjureEvalVisual",
-    },
+    keys = { "<localleader>r" },
     -- [Optional] cmp-conjure for cmp
     dependencies = {
         {
--- a/README.adoc
+++ b/README.adoc
@@ -82,6 +82,81 @@ use 'Olical/conjure'
 Plug 'Olical/conjure'
 ----
 
+=== https://github.com/folke/lazy.nvim[lazy.nvim]
+
+[source,lua]
+----
+return { "Olical/conjure" }
+----
+
+To take advantage of lazy loading, a bit more involved configuration is required:
+[%collapsible]
+====
+[source,lua]
+----
+return {
+    "Olical/conjure",
+    cmd = {
+        "ConjureEval",
+        "ConjureSchool",
+        "ConjureConnect",
+        "ConjureClientState",
+        --
+        "ConjureLogSplit",
+        "ConjureLogVSplit",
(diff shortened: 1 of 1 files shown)


Olical:
Yeesh that's a lot of extra config! 😅 Hopefully I won't need to maintain those instructions too often 😬 I've been vocal about it before elsewhere, but I think lazy loading for the most part is a waste of time unless the plugin consists of many slow VimL files and takes 100ms+ on it's own to load. My Neovim loads in 150ms with zero lazy loading, personally, I think that's absolutely fine. I _might_ one day time box 30 minutes to see if there's some low hanging fruit I should disable or lazy-fy in my AstroNvim config, but not right now.

Conjure already lazy loads all of it's components as they're required through Aniseed's `autoload` which replaces the Lua `require` with a lazy alternative, but it does still do some work at startup to create autocmds. Just not much in my opinion.

Are you seeing significant differences when lazy loading? And how often do you enter, say, a Clojure buffer and NOT want Conjure to load?

Maybe the better approach is just to load on any Conjure supported filetypes, that way your default `nvim` command doesn't load it but as soon as you enter a file for something supported it incurs the < 10ms (I think on almost all machines? Maybe I'm wrong!) cost of loading a couple of initial Lua files.

This feels a little too fiddly and overkill for my liking. I don't like that I, and other users, would need to maintain that list of commands etc. I just wonder how we could get 99% of the value with a fraction of the config.

Interested in hearing your experiences with how much lazy loading helps with Conjure and if just loading on filetypes would be acceptable too? That way users could drop filetypes they're not interested in and add the ones they care about. Unlike the command list which they have to keep in sync with development.

An alternative might be to do a little more optimising of Conjure's startup so it does even less on first load and does nothing until you actually enter a supported filetype.

Interested in your thoughts on this! I know it's more than you bargained for with a readme update, I just see a lot of people trying to lazy load _everything_ and I'm interested in why/what motivates it/what could help. Because I'm clearly not the right demographic with my opinions of "meh, good enough" 😄 I need some outside opinions to help shape these docs and potential Conjure changes.

Olical:
If I can't get Conjure lazy enough for people out of the box then maybe the alternative change would be to add a function users can call when it _is_ lazy loaded. That way at least the path to loading it later is paved rather than something adhoc that could be broken by some rewrite of something inconspicuous. A well named function that is obviously for users to call would protect that method into the future for people.

p00f:
I don't lazy-load much either but conjure makes it go from 58 to 65ms

Quitlox:
> Interested in your thoughts on this! I know it's more than you bargained for with a readme update, I just see a lot of people trying to lazy load everything and I'm interested in why/what motivates it/what could help.

No problem, I understand that adding such a large configuration is undesirable from a maintenance perspective.

I'll give some of my thoughts about this change. First of all, part of my motivation was just trying to see how far I could get with lazy loading my config :stuck_out_tongue:, but it's not the only motivation.

Because of reasons I sometimes have to work with Neovim inside of Powershell on Windows, and its terribly slow. Once it has booted, its doable but Neovim takes roughly 6 seconds to start. Every speed-up here helps, though admittedly, even on that setup conjure only contributes about ~30ms. 

Secondly, for my workflow I really don't need conjure most of the time. I only use conjure as my REPL for Python. I work a lot in Python, but only every once in a while require a REPL. So lazy loading on filetype wouldn't make a lot of sense for me either.

Having said all that, I realize the config can be simplified a bit by removing the `cmd` section (unless you want to use a conjure command before having activated it via a keymap). See last commit.

Olical:
> its doable but Neovim takes roughly 6 seconds to start.

Ouuuch, sorry to hear that! I'll give this a little more thought, I feel like in your semi-special case situation the best approach would be to have a custom mapping that triggers the load.

So for most users, they should just load on ft (which doesn't even work with conjure without calling a fn because after the ft it's too late for it's autocmds to fire!) but for some they should add a manual switch that loads it. Maybe improving the ft based lazy load experience so that _just works_ will help the lazy config for most users look quite nice.

For your case I'm still leaning towards suggesting a custom mapping or command or maintaining a list of _some_ mappings you know you personally use to load everything. Or a custom mapping that has nothing to do with Conjure that loads it that you define.

Olical:
Merged to `develop` with a couple of tweaks, I'm going to recommend loading on filetypes, not keys for now. I'll do some work to make lazy loading work more seamlessly without the special function calls etc if possible. And hopefully provide a nice hook for you too! I think the config you suggested is great for you, but I don't want to modify the default prefix for other users in docs about lazy loading if possible. I'll have a noodle about it.

Olical:
Thanks for the work on this! And thanks for prompting more thought from me.

Issue #432, opened 2022-10-19
client for psql (postgresql repl) 


AbhinavOmprakash:
Hi, I was wondering if it would be possible to add [psql](https://www.postgresql.org/docs/current/app-psql.html) client support for conjure. 

I use psql extensively and conjure as well, and I would like to work on this if you give me some pointers. 

I know that tree-sitter can be used to parse the syntax, and I can look at the python implementation for pointers. 

I'm not certain how to connect conjure with the psql client and pass data between them. (whether it will be remote or via stdio) 

Also, in order to start the psql client, a database URL must be given to it. This can be added to the user's init file, I'm not sure how to pick it up from conjure. (I bring this up because this will directly impact the UX)

Invertisment:
Clojure plugin loads the nREPL port from a file in the working directory.
That means that the plugin loads only after you enter a file.
And that also means that you can have a config file in your project that you could use to read this DB URL from.

Conjure can also switch between REPL connections (I think) so you may want to look into the initialization code of Clojure.

AbhinavOmprakash:
That is an idea. In the case of clojure you have deps.edn/project.clj files which contain the port information. In the case of db connections there is no such standard file that I'm aware of. 

a `.psqlrc` file does exist, but that's just config specific a psql application. Any ideas?

Invertisment:
Read it from and environment variable or hardcode into your `.vimrc`. Just do something simple, don't spend time on it. And then later you'll decide what you want.

AbhinavOmprakash:
okay, sure. 
also, how to connect conjure with the psql client and pass data between them?

Olical:
That would have to be over stdio, there's already a shared module all other stdio clients use to spin up a sub-process and send/receive data. That module needs a rewrite / sequel that works a lot better, but the existing one is good enough for a bunch of simple cases.

Olical:
This can probably be a copy/paste/modify job of an existing stdio client, of which there are quite a few!

AbhinavOmprakash:
Alright. Thanks for the pointers. Let me try to implement something :)

russtoku:
[...]

PR #511, opened 2023-06-25
Add client for SQL

russtoku:
A first attempt at a client for interacting with an SQL REPL to satisfy issue #432.

Although built using PostgresQL's `psql` command line utility, it should work with MySQL's `mysql`, Oracle's `oci`, and others. But I did not try any other.


russtoku:
I'm still unsure about how the `next-in-queue` function plays into the REPL interaction. I copied `remote/stdio.fnl` to `remote/stdio-rt.fnl` to get the responses from the `psql` REPL displayed in the Conjure log buffer. I couldn't figure out why I don't see the `psql` prompt in the responses from the `on-stdout` or `on-stderr` callbacks.

So, this is more of a proof-of-concept thing.

russtoku:
Doh! `psql` doesn't send a prompt when not in an interactive terminal.

russtoku:
Added a SQL statement with meta-command for psql to demo that `<localleader>ei` (interrupt REPL) stops the meta-command and doesn't kill the REPL. Updated `conjure-client-sql-stdio` doc to reflect this.

Also, noticed that when stopping the REPL, the REPL process became a zombie (half-dead -- you can find it with `ps` but it's non-functional). It looks like just calling `process_kiill` without `handle:close` will clean up resourses automatically and lets the REPL process die normally.

Olical:
This looks wonderful! You caught everything I think, all the right docs, config and files. Amazing!

And great catch on the zombie process, that'd be easy to miss. The `next-in-queue` system I think was a misguided attempt at making evals more robust and separated from user interactions, so things could be queued up by the user or the system and then executed as quickly as possible by some other sub-system.

I think it didn't really play out how I wanted so now it's tech debt. I'm okay with the stdio-rt copy for now but only really because I want to completely rewrite stdio which would hopefully supersede stdio+stdio-rt, supporting everything we need from both in a nicer way. Again I refer to #500 and something I need to do eventually 🥲 

For now I'm focussing on some Fennel tooling improvements to make the development nicer, but I'll get to that eventually.

I'm going to merge this to develop to get it in people's hands, once again, amazing work and thank you!

tssm:
Hey @russtoku, I just wanna thank you for this! I can confirm it works with MySQL ❤️

russtoku:
@tssm , that's good news! Glad it works for you!

PR #512, opened 2023-07-02
docs: remove g:conjure#mapping#eval_comment_form (unused)

MisterFISHUP:
Hi,

I noticed that the mapping `g:conjure#mapping#eval_comment_form`, which defaults to `ec`, doesn't really exist. I've checked the source code and it seems that it's not used anywhere, so I think we can safely remove it from the doc to avoid any confusion.

Thank you for creating and maintaining this amazing project!

### DIFF
--- a/doc/conjure.txt
+++ b/doc/conjure.txt
@@ -486,12 +486,6 @@ configuring these values from Lua it's just `true` and `false`.
             Evaluates the form at the marks location.
             Default: `"em"`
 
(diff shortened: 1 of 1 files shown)


Olical:
Hah, yeah, this was replaced by three mappings to eval the current form, root form and word under cursor and store the result in a comment. I guess I forgot to remove this one, thanks!

Olical (approved):

Olical:
Merged to develop

Issue #466, opened 2023-02-23
Common Lisp Completion


LordMZTE:
It'd be awesome to get Conjure to provide completion for Common Lisp. I've looked far and wide, and couldn't find any suitable language server or other completion provider that's easy to integrate into nvim-cmp. Considering that Conjure otherwise makes for an amazing development environment, it'd be great to expand it further.

glyh:
I'm not sure about this. In my perspective, conjure is designed as an evaluation environment, rather than a completion source.

Also does that require common lisp codes in this project? 

What's preventing you from writing a common lisp language server? That feels more modular than just integrate the common lisp's functionality into this project.

Olical:
I'm very strongly leaning towards "just use LSP" for anything like this these days. The tooling outside and inside of Neovim has come along so far. LSP + DAP cover 90% of what everyone needs beautifully, Conjure should narrow it's focus down onto running your program and executing chunks of code, allowing you to merge the concept of editor and REPL. I'm going to close this in favour of recommending LSP for now, feel free to reopen or comment with opinions against this!

LordMZTE:
This isn't an option, mainly since I've not found an LSP that works, but also because there is not really a way to do an LSP in the traditional sense in Common Lisp. The whole language environment is so dynamic that static analysis like what an LSP would typically do is impossible. The only way an LSP in CL could possibly work is if it were running under the same runtime as the REPL server Conjure is connected to. However, such an LSP does not exist.

Considering that this would then mean having 2 connections between the CL runtime and Neovim (1 for Conjure, 1 for this **hypothetical** LSP) whereas the SWANK API Conjure uses already provides LSP-like APIs for completion and such (I believe?), this solution makes little to no sense.

Olical:
Ah interesting, I didn't realise static analysis was kinda not valid in the CL space, I've never looked into it too closely and have very surface level knowledge. Thanks for explaining! I'll reopen and keep it in the backlog of things to enhance if there's no LSP with autocomplete for CL, I just thought it'd be similar to Clojure or something where yes there's REPL state, but it should almost perfectly reflect the static analysis anyway.

glyh:
Well then the codebase should be shifted to another project that implement LSP based on Sly/Swank or whatever. BTW, Olical says he will likely to redesign this plugin (reference [here](https://github.com/Olical/conjure/issues/433#issuecomment-1621378084) ). I guess by then people can implement whatever they want in a separate plugin rather than keep throwing everything together.

[...]

PR #517, opened 2023-07-20
Add completions to Common Lisp client

russtoku:
I hope this helps a little to resolve issue #466.

Some issues:
- The candidates are identified as *TEXT*. I haven't figured out how to at least annotate them as *symbols* or something that would indicate that they are from from Swank.
- The candidate list returned from Swank is based the first letter typed. 'lo' and 'la' would share the same list of candidates.
- If you trigger omnicompletion after typing say 'lo', you'll get an error message but the completion pop-up window is not affected.

Suggestions welcome.


LordMZTE:
This is awesome! Apart from what you stated already, I have 3 more criticizms:
- The Conjure output window is spammed a little during completion. It'd be great if output from completions wasn't sent to it.
- An error occurs when completing a package, for example triggering a completion here: `uiop:|` (where `|` represents the cursor) causes this error to appear:
```
Error executing vim.schedule lua callback: ...ker/opt/conjure/lua/conjure/client/common-lisp/swank.lua:173: too many results to unpack
stack traceback:
	[C]: in function 'unpack'
	...ker/opt/conjure/lua/conjure/client/common-lisp/swank.lua:173: in function 'dispatch'
	...ker/opt/conjure/lua/conjure/client/common-lisp/swank.lua:213: in function 'parse_separated_list'
	...ker/opt/conjure/lua/conjure/client/common-lisp/swank.lua:232: in function 'parse_result'
	...ker/opt/conjure/lua/conjure/client/common-lisp/swank.lua:249: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>
```
- This may be a limitation with swank, but currently, completions are shown which won't work due to being declared in another package. For example, your code would suggest `run-program` whereas that function needs to be refered to by `uiop:run-program`.

Thanks for you great work! This is really good already!

Olical:
Tried it locally and I got some completions! I had to trigger omnicomplete though with `<c-x><c-o>`  which I found odd.

Olical:
A fun error if I trigger omnicompletion with no text on the line:
image

I _think_ there's an unpack somewhere in our lua and it's being asked to unpack too many values. I didn't know there was a limit so The More You Know! I'll see if I can find it.

Olical:
I've replaced the unpack in question with `(str.join (a.map string.char stack))` and it can now handle any size. So we should avoid using `unpack` for unbounded tables or at least tables we think could get very large. Merged to develop!

russtoku:
Yes, that `unpack` was the cause of the error. Thanks for that fix.
I'm still trying to work through some of the short-comings mentioned but I'm glad we've gotten thing this far.

PR #520, opened 2023-07-28
Allow overriding opts context in eval-str

julienvincent:
## What

Allow the caller of `eval-str` to explicitly specify the execution context (`opts.context`) instead of having it always be replaced by the `namespace` of the active buffer.

## Why

I've setup for myself some global keybindings I can run from any buffer to eval code in predefined namespaces such as `user`.

Here is an example of one such mapping:

```lua
{
  "<leader>§",
  function()
    local fn = require("conjure.eval")["eval-str"]

    client["with-filetype"]("clojure", fn, {
      origin = "julienvincent.custom-command",
      code = "(do (ns user) (tap> (reset)))",
    })

  end,
  desc = "user/reset",
},
```

This kinda works except that the auto-extracted `opts.context` can mess with this operation. An example would be if I were to be in the following situation:

- I have a freshly started clojure repl and;
- I have a clojure namespace open in the active buffer and;
- I execute this mapping

Then the namespace of the buffer I have open will be evaled by conjure (`(ns a.b.c)`) and so when my (reset) function fires it will fail as it sees my namespace has been loaded but none of the symbols in the namespace are loaded. So I would then need to eval buffer and try again.

This is a bit of a convoluted example, but in my workflow this can be quite annoying. The main point is that the context of programatic `eval-str` calls should be controllable by the user/developer.

I worked around this issue for a while by using the `conjure.client.clojure.nrepl.action` namespace directly:

```lua
...
local fn = require("conjure.client.clojure.nrepl.action")["eval-str"]
...
```

But this doesn't get the goodies from the higher-level `conjure.eval` implementation, which I want :)

## Result

After this PR I should be able to do the following in my global mapping:

```lua
{
  "<leader>§",
  function()
    local fn = require("conjure.eval")["eval-str"]

    client["with-filetype"]("clojure", fn, {
      origin = "julienvincent.custom-command",
      context = "user",
      code = "(do (tap> (reset)))",
    })

  end,
  desc = "user/reset",
},
```

And everything works as expected.


Olical:
LGTM, goo explanation and implementation, thanks!

PR #536, opened 2023-10-27
Add a stdio based Janet client

sogaiu:
This is a basic stdio client for Janet based on some of the existing stdio clients.

There is already a nice netrepl-based client for Janet, though that requires one to install and setup a netrepl server for use.

A stdio-based client allows one to avoid the installation of both `jpm` and `spork` (possibly a speed bump for folks who are in the process of evaluating [1] Janet) but also might be nice for quick interactive sessions.  As with other stdio-based solutions though, functionality is reduced.  Perhaps a worthwhile trade-off for some situations.

This PR includes document updates and additions but the Wiki has not been updated yet.

---

[1] Pun intended :P


sogaiu:
Sorry, I failed to choose the `develop` branch when creating this PR initially.  Have made an attempt to correct that.

Olical:
This is really cool! Just tried it out and it worked great for me! I'd even consider making this the default because I love the "it just works" approach but that might break some users' workflows, so maybe not just yet. Thank you so much for putting all of the effort into this and I'm sorry it's taken so long to merge it, I've just been really busy with life, work and OSS, I must've lost this notification in my inbox 😢 

Merged now though! Will release ASAP as soon as some other issues are confirmed to be fixed / addressed by other people.

Issue #412, opened 2022-09-11
GNU Guile Scheme REPL not showing output


francium:
I'm trying to do some very simple Scheme REPL work.

I've got a very simple config and almost everything is working except for,
```
(display "Hello, World!")
```

Which is just giving me, when I do `<leader>ee`,
```
; --------------------------------------------------------------------------------
; eval (current-form): (display "Hello, World!")
; Empty result
```

Here's my relevant config,
```
Plug 'https://github.com/Olical/conjure'
let g:conjure#filetype#scheme = "conjure.client.guile.socket"
let g:conjure#client#guile#socket#pipename = ".guile-repl.socket"
```

And the script for creating the REPL and socket,
```
#!/usr/bin/env bash

SOCKET=$(git rev-parse --show-toplevel 2> /dev/null || pwd)/.guile-repl.socket

echo "$(tput bold)$SOCKET$(tput sgr0)"

if [ -f $SOCKET ]; then rm $SOCKET; fi
guile --listen=$SOCKET

rm $SOCKET
```

In the `guile` command line REPL, I get the expected result,
```
scheme@(guile-user)> (display "Hello, World!")
Hello, World!scheme@(guile-user)> 
```

Contrast this with `ConjureSchool`,
```
  (print "Hello, World!")
```
Here doing `<leader>ee` on this form results in,
```
; --------------------------------------------------------------------------------
; eval (current-form): (print "Hello, World!")
; (out) Hello, World!
nil
```
The `(out Hello, World!` as expected.

So is this perhaps a GNU Guile limitation?

I'm very new to Lisp, Scheme, Guile and Conjure, so hopefully I'm not making some very obvious novice mistake. If I am please let me know.

Olical:
I think you need a newline at the end of the display call? Which isn't ideal, but that should work?

This is something that would probably be fixed by the future stdio client support rewrite.

francium:
So this right,
```scheme
(display "Hello, World!\n")
```
Then `<leader>ee` gives,
```
; --------------------------------------------------------------------------------
; eval (current-form): (display "Hello, World!\n")
; Empty result
```

francium:
I tried out `vim-sline` with this config,
```
Plug 'https://github.com/jpalardy/vim-slime'
let g:slime_target = "tmux"
let g:slime_paste_file = tempname()
```
That was working without any issues, so may just be an issue with with stdio client in conjure as you mentioned.

Olical:
[...]

PR #552, opened 2024-01-21
guile: append stray output to log

austinmlv:
This checks to see if there was a result found and if so gathers all text from after the end of the last prompt until `$n = ...`. If not then everything from the last prompt up until the next is gathered. In either scenario the stray text is appended to Conjure's log. This makes the Guile client more consistent with the built-in Fennel client.

Also, this uses Fennel's support for multiple cond/body pairs within if forms thus removing a nested if.

This is a possible fix for #412


Olical:
Oh wonderful! Thank you! Taking a look now.

Olical:
Tested locally and it's working, however as you said in the original issue it doesn't cover muti value returns. I think we might need a bit of a rethink about how this works if we want to address all of the issues. This is a good incremental improvement though. Thanks!

Issue #555, opened 2024-02-07
Common-Lisp: Only first expression is evaluated using ConjureEvalBuf


iicurtis:
When using the buffer or visual select evaluations, only the first expression is evaluated.

Neovim Config: https://github.com/LazyVim/starter
with `lua/plugins/example.lua` updated to 

```lua
return {
  {
    "Olical/conjure"
  },
}
```
Example file:
```lisp
(defun hellofun ()
           (write-line "Hello, World"))

(hellofun)
```
Expected Ouput:
```lisp
; eval (buf): ...src/test.lisp
; "Hello, World"
```

Actual Output:
```lisp
; eval (buf): ...src/test.lisp
HELLOFUN
```

Olical:
This might just be an issue with capturing the output from the REPL, I have
a hunch that the code IS evaluating, we're just not capturing the results
from the REPL properly in this scenario.

iicurtis:
I'm not sure that's the case. For example, if I run a larger example file like the one below, none of the functions are defined if I try to evaluate current expression with `<leader>ee`

```common-lisp
;;;; Day 1: Calorie Counting
(in-package #:advent-of-code-2022)

(defparameter *day1-input* #P"../inputs/day01.txt")

(defun run-day1 ()
  (let ((input (uiop:read-file-lines *day1-input*)))
    (time (format t "Part 1: ~a~%" (day01-part1 input)))
    (time (format t "Part 2: ~a~%" (day01-part2 input)))))

(defun day01-part1 (input)
  (reduce #'max (read-cals input)))

(defun day01-part2 (input)
  (reduce #'+ (read-cals input) :end 3))

(defun read-cals (input)
  "Convert input into groups of calorie sums"
  (let ((cals (sort
                (mapcar (lambda (nums) (reduce #'+ (mapcar #'parse-integer nums)))
                        (split-sequence:split-sequence "" input :test #'equal))
                #'>)))
      cals))

(run-day1)
```

Evaluating `(run-day1)` results in 
```common-lisp
; --------------------------------------------------------------------------------
; 127.0.0.1:4005 (connected)
; --------------------------------------------------------------------------------
; eval (buf): ...s/advent-of-code/2022/lisp/src/day01.lisp
#<PACKAGE "ADVENT-OF-CODE-2022">
; --------------------------------------------------------------------------------
; eval (current-form): (run-day1)
; The function COMMON-LISP-USER::RUN-DAY1 is undefined.
```

[...]

PR #556, opened 2024-02-09
Common-Lisp: Use interactive-eval-region

iicurtis:
**Disclaimer:** I have absolutely no idea what I'm doing. I captured some logs from emacs' SLY with wireshark and noticed that SLY uses `interactive-eval-region` for its buffer evaluation while we're using `eval-and-grab-output`. I have no idea why, but it seems to fix the ConjureEvalBuf only loading the first expression seen in #555. 

The swank source is totally opaque to me, but someone else may find it useful: https://github.com/slime/slime/blob/master/swank.lisp.

Prefer interactive-eval-region over eval-and-grab-output, as the latter requires list output parsing.

### DIFF
--- a/fnl/conjure/client/common-lisp/swank.fnl
+++ b/fnl/conjure/client/common-lisp/swank.fnl
@@ -87,7 +87,7 @@
         (remote.send
           conn
           (str.join
-            ["(:emacs-rex (swank:eval-and-grab-output \""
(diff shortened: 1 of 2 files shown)


russtoku:
Seems to work. Thanks!

iicurtis:
Closing this pull request until I can find a way to get both the stdout and the evaluation output.

PR #560, opened 2024-02-27
Debugger: Show value of :debug-value if present

dradtke:
When using the debugger, results from `:ConjureCljDebugInput eval` were not being displayed in the log buffer. It looks like these results are returned in the `:debug-value` field, so Conjure just needs to check if that field is available and print it if so.

For comparison, [here](https://github.com/clojure-emacs/cider-nrepl/blob/3824d72f9ba1a14f9ce9910500db0885f6b39c23/src/cider/nrepl/middleware/debug.clj#L390) is where the result is read in `cider-nrepl`, confirming that this is indeed the right field.

### DIFF
--- a/fnl/conjure/client/clojure/nrepl/debugger.fnl
+++ b/fnl/conjure/client/clojure/nrepl/debugger.fnl
@@ -65,6 +65,9 @@
         {})
       {}))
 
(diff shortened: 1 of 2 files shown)


Olical:
This looks good, thank you! I'll give it a go on develop tomorrow.

PR #563, opened 2024-03-11
Fix doc-str Guile

DMClVG:
doc-str would silently fail and not display anything in nvim when used with Guile due to a missing check on the optional opts.on-result property. This PR fixes it


Olical (approved):
Thanks! This looks like a good fix. I would use a `when` instead of an `if` when it's a single branch though, I'll correct that post-merge happily because this is still fine Fennel, just would fail linting if such linting existed. (maybe it does!)

DMClVG:
You're right. Force of habit lol

PR #569, opened 2024-04-16
Clojure: Pretty-print test failures

dradtke:
I am often running Clojure tests that compare two large sets of data, and Conjure currently doesn't make it easy to see what failed in these scenarios. Here's a simple test to demonstrate:

```clojure
(deftest compare-alphabets
  (let [a1 {:a 1 :b 2 :c 3 :d 4 :e 5 :f 6 :g 7 :h 8 :i 9 :j 10 :k 11 :l 12 :m 13 :n 14 :o 15 :p 16 :q 17 :r 18 :s 19 :t 20 :u 21 :v 22 :w 23 :x 24 :y 25 :z 26}
        a2 {:a 1 :b 2 :c 3 :d 4 :e 5 :f 6 :g 7 :h 8 :i 9 :j 10 :k 11 :l 12 :m 13 :n 14 :o 15 :p 16 :q 17 :r 18 :s 19 :t 20 :u 21 :v 22 :w 23 :x 24 :y 25 :z 27}]
    (is (= a1 a2))))
```

When this test is run in Conjure, it prints a result like this to the buffer log:

```clojure
; expected: (= a1 a2)                                                                                                                                                                                                                                                                                                                                                                                                                  
;   actual: (not (= {:y 25, :q 17, :r 18, :v 22, :o 15, :n 14, :w 23, :m 13, :e 5, :s 19, :l 12, :k 11, :z 26, :g 7, :c 3, :j 10, :h 8, :b 2, :d 4, :f 6, :t 20, :x 24, :p 16, :i 9, :a 1, :u 21} {:y 25, :q 17, :r 18, :v 22, :o 15, :n 14, :w 23, :m 13, :e 5, :s 19, :l 12, :k 11, :z 27, :g 7, :c 3, :j 10, :h 8, :b 2, :d 4, :f 6, :t 20, :x 24, :p 16, :i 9, :a 1, :u 21}))
```

Naturally, it's quite difficult to determine _what_ actually failed here.

With the changes included in this PR, the result instead looks like this:

```clojure
; - {:z 26}                                                                                                                                                                                                                                          
; + {:z 27}
```

This output is inspired by the recommendations for [Go](https://go.dev/wiki/TestComments#print-diffs), and is a clear and concise way to see the cause of the test failure.

## How it Works

My first observation was that Conjure does not pretty-print these failures, even though it will pretty-print direct evaluations of these structures just fine. I believe that is because Clojure's default test reporting function uses `pr-str` directly [here](https://github.com/clojure/clojure/blob/4b9eadccce2aaf97e64bcc8e35c05c529df8fdd2/src/clj/clojure/test.clj#L381-L382), rather than a configurable printing function. (Interestingly, Calva does seem to do basic pretty-printing of test failures, though I wasn't able to determine exactly how it works)

My solution here is to override the reporting multimethod with an implementation that is very similar, but checks specifically for the `actual` case of `(not (= a b))`. If this case is detected, then `clojure.data/diff` is used to diff `a` and `b`, and the results are pretty-printed using the format shown above.

I'm pretty happy with the end result on my machine, but understand that this implementation as-is may not be ideal for everyone. So an alternative may be to allow users to inject their own `clojure.test/report` overrides, or to provide some other way to make the test-reporting behavior a little bit more configurable.


Olical:
Thank you for the contribution! I'm assessing it now, I agree that it won't be right for everyone and it's the kind of thing I'll probably stick behind a flag because some people will be upset if their test workflow (even if it's just the diff output) changes from under them. Not sure if I'll default this to on or not yet, I'll give it a go.

My worry is that some data types will cause errors in the diff function, can it handle a `LocalDate` or some other random Java classes for example? Maybe it can! I just need to try that out to see if there are any edge cases that could cause this to blow up.

Personally, Kaocha's output is my favourite. Could be cool to be able to piggieback on whatever Kaocha does if it's on the classpath since it's extremely field tested. There may also be other nREPL APIs I should be using to execute tests that would return data in a nicer shape than what we currently have.

Olical:
For context, here's how Kaocha failures look for me

image

Olical:
So far the only thing I'm concerned about is:

 - Does it handle all data types? Are there edge cases that could throw exceptions?
 - It doesn't show the full expected and actual values which although verbose and hard to read on one line, are sometimes very useful for spotting exactly what went wrong. So I might tweak this so instead of an either/or change, this supplements the main output.

Olical:
Oh I see you were checking if things are sequences before diffing them, that should be safe. And what do you think of this output? I've turned it into pretty printing the full data with a diff summary to follow:

```clojure
; run-ns-tests: clojure-dap.server-test
; 
; Testing clojure-dap.server-test
; 
; FAIL in (run) (server_test.clj:37)
; multiple inputs generate multiple outputs
; expected: (=
;  [{:body handler/initialised-response-body,
;    :command "initialize",
;    :request_seq 2,
;    :seq protocol/seq-placeholder,
;    :success true,
;    :type "response"}
;   {:event "initialized", :seq protocol/seq-placeholder, :type "event"}
;   {:body {},
;    :command "attach",
;    :request_seq 2,
;    :seq protocol/seq-placeholder,
;    :success true,
;    :type "response"}]
;  (vec (s/stream->seq output-stream)))
;   actual: (not
;  (=
;   [{:body
;     {:supportsCancelRequest false,
;      :supportsConfigurationDoneRequest true},
;     :command "initialize",
;     :request_seq 2,
;     :seq -1,
;     :success true,
;     :type "response"}
;    {:event "initialized", :seq -1, :type "event"}
;    {:body {},
;     :command "attach",
;     :request_seq 2,
;     :seq -1,
;     :success true,
;     :type "response"}]
;   [{:type "response",
;     :command "initialize",
;     :seq -1,
;     :request_seq 1,
;     :success true,
;     :body
;     {:supportsCancelRequest false,
;      :supportsConfigurationDoneRequest true}}
;    {:type "event", :event "initialized", :seq -1}
;    {:type "response",
;     :command "attach",
;     :seq -1,
;     :request_seq 2,
;     :success true,
;     :body {}}]))
; -  [{:request_seq 2}]
; +  [{:request_seq 1}]
```

In my opinion, this is now all the benefits with none of the controversy / drawbacks that would require configuration options to keep everyone happy. I _think_ everyone should be happy with this?

I also really need to change how this Clojure preamble is stored, it needs to be in a `.clj` that we read in the plugin somehow, but finding the file on disk will be tricky...

Editing Clojure inside Fennel strings is a nightmare and I'm sorry you had to deal with it!

Olical:
Merged to develop wit the modifications I mentioned above, please let me know what you think and thank you for this improvement!

philomates:
hello :wave:
Cool idea that seems quite useful.
I'd love to use something like this but also realized that it isn't compatible with https://github.com/nubank/matcher-combinators which I use heavily for asserting nested data structures + pretty printing mismatches.
I noticed this when I recently switched to the `develop` branch for some other changes and found that my test output changed.

test failures using `match?` from https://github.com/nubank/matcher-combinators will return the original datastructure with some mismatch annotations that have their own pretty print dispatch https://github.com/nubank/matcher-combinators/blob/beae388ac30d4079a68619aba36ff7e152b20e42/src/clj/matcher_combinators/clj_test.clj#L155-L157. 

So normally I would see something like
2024_05_31_12h22m43s

But after the change in this PR it starts looking like
2024_05_31_12h25m24s


I'll go back to the main branch for now, but would be nice to have this behind a feature flag before it gets into main so I can toggle the functionality.

Olical:
Ah I was worried something like this might happen! I'll look at making it more flexible or at least putting it behind a flag for you.

Olical:
Think I've made it so it works for both cases now @philomates 😄 let me know if you run into more issues though!

Olical:
I think `prn` is better than `pprint` in this case anyway so the expected and actual fit on one line. It's also the change that makes sure the matcher combinator printer kicks in. It was being bypassed by pprint basically.

philomates:
gave it a quick try and looks good; thank you!

PR #595, opened 2024-08-19
Convert extract tests, fix test setup

russtoku:
This is my first try at converting an Aniseed test to *busted + luassert*.

I picked one that required a supporting test module that appears to be a one-off thing. I hope the integration of that module is OK. fennel code formatting might be a bit messed up by my Neovim set up. Advice and corrections are more than welcome.

### DIFF
--- a/fnl/conjure-spec/extract_spec.fnl
+++ b/fnl/conjure-spec/extract_spec.fnl
@@ -74,7 +74,7 @@
         (it "matching nothing"
           (fn []
             (at [2 0])
-            (assert.equal nil (extract.form {}))))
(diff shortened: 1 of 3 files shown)


russtoku:
Sorry, minor fix up to follow https://lunarmodules.github.io/busted/.

Olical:
Nice! Thank you very much for this. I will probably try to recover that buffer helpers module, I think it gets used in a few places so it'll be worth converting that too.

Thank you for helping out 🎉

Olical:
Oh after re-reading I realise that it was a one off and see why you
integrated it. Oh well, we have a generic test util module now for the
future if required.

PR #596, opened 2024-08-20
Convert uuid.fnl and add test

russtoku:
There are commented functions at the bottom of uuid.fnl and are useful when using Conjure for development.

To make things externally tesable, I exported the formerly private cats-and-dogs list to use in a busted test to verify that `v4` returns something in that list. I'm not sure if it's a good idea so if you think that we want to keep the cats-and-dogs private, I'm OK with that.


Olical (approved):
Making that public seems fine to me!

PR #597, opened 2024-08-21
Convert tree-sitter.fnl, timer.fnl, text.fnl, stack.fnl, and tests

russtoku:
More files converted along with corresponding tests.


Olical:
Looks lovely! Thank you very much! I really like how much smaller every `.lua` file is after simplifying it and removing the Aniseed macros. They didn't add much bloat, but they weren't free. I wonder how much that added up in total since Conjure is comprised of a lot of code and files, will we see a noticeable improvement to startup times on slow machines I wonder. Conjure has always been a negligible impact to my Neovim startup, but if some have struggled on slower drives or CPUs I hope this makes a difference for them.

PR #599, opened 2024-08-27
Convert SQL client

russtoku:
Tested  SQL client by evaluating SQL statements. The client uses a modified stdio module.


Olical:
Another chunk done! Thank you!

Olical:
Tried it out and it works for me too!

PR #600, opened 2024-08-28
More conversions

russtoku:
Converted clients:
- snd-s7
- Python (with small refinements)
- Scheme
- Hy (not tested)
- Julia (not tested)
- Janet stdio (not tested)
- Fennel stdio

Converted:
- remote/transport/base64
- remote/transport/bencode
- remote/stdio and associated test

Touch up:
- dev/python/sandbox.py
- remote/stdio-rt

### DIFF
--- a/dev/python/sandbox.py
+++ b/dev/python/sandbox.py
@@ -50,6 +50,7 @@ def fn_with_multiline_str():
 fn_with_multiline_str()
 
 
+"""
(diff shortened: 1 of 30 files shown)


Olical (approved):
🚀  look at those deletions! I love to see the code shrink while everything gets easier to follow and work with. This is so good. Thanks as always!

PR #601, opened 2024-09-01
More nfnl conversion

russtoku:
- Converted:
  - Common Lisp client
  - remote/stdio2.fnl
  - conjure.editor
  - conjure.dynamic
  - extract/searchpair
  - client/fennel/aniseed
- Added busted tests:
  - remote/stdio2_spec.fnl
  - client/fennel/aniseed_spec.fnl
- Touched up Racket client (converted nvim.fn to vim.fn)


russtoku:
Apologies, I pushed my local commit for the fennel/aniseed client too soon. It was supposed to be for the next batch.

Unfortunately, the Circle CI build is failing on the `lua/conjure-spec/client_spec.lua` test now. It looks like the ordering of the items returned by `client.each-loaded-client` is not what is expected.

russtoku:
Not sure if it's the *right* thing to do but I sorted the sequences being compared in the "each-loaded-client runs a function for each loaded client" test in `client_spec.fnl` to fix the test.

Olical:
Sorting sounds like the right move! I did think that could be flaky at the time, I should've written a better test in the first place. Great catch! Looking at merging this now and will try to fix #602 which I probably caused.

Olical (approved):
🎉

Olical:
Found an error when trying to autocomplete which is interesting, I think only post merge? Although maybe it was there before. But don't worry, I'm on it 😄

Olical:
Ah! Here! The Fennel LSP really helps with spotting these undefined references.
image

I'll get that fixed, no big deal. Just need to make sure any references to magic Aniseed values are removed or replaced in some way during the conversion.

PR #618, opened 2024-11-17
Improve evaluation of class, decorated class, and decorated function definitions in Python with examples

russtoku:
Improved evaluation with `<localleader>ee` for class definitions and decorated class and function definitions in the Python client. Added some representative code for these cases to `sandbox.py`.


fnl/conjure/client/python/stdio.fnl
```diff
@@ -38,13 +38,21 @@
...
   [node]
-  (or (= "expression_statement" (node:type))
-      (= "import_statement" (node:type))
-      (= "import_from_statement" (node:type))
-      (= "with_statement" (node:type))
-      (= "function_definition" (node:type))
-      (= "for_statement" (node:type))
-      (= "call" (node:type))))
+  (log.dbg "form-node?: node:type =" (node:type))
+  (log.dbg "form-node?: node:parent =" (node:parent))
```
Olical, line 42: Ohh good idea for a debug line, this will help diagnose some things.

fnl/conjure/client/python/stdio.fnl
```diff
@@ -38,13 +38,21 @@
...
+        (= "import_from_statement" (node:type)) true
+        (= "with_statement" (node:type)) true
+        (= "decorated_definition" (node:type)) true
+        (= "for_statement" (node:type)) true
+        (= "call" (node:type)) true
+        (and (= "class_definition" (node:type))
+              (not (= "decorated_definition" (parent:type)))) true
+        (and (= "function_definition" (node:type))
+             (not (= "decorated_definition" (parent:type)))) true
+        false)))
```
Olical, line 55: Nice! I assume this is inspired by the Julia client changes? Discovering that `:InspectTree` thing has made this a lot easier to reason about thankfully.

Not sure if we explicitly need to return true or false here though, I feel like we can get away with truthy and falsy. But if this is easier to reason about or you had issues without explicit boolean return values then this is fine.

Olical (approved):

Olical:
Tried it out locally and it works wonderfully, thanks! This tree sitter system is working pretty well 😄

fnl/conjure/client/python/stdio.fnl
russtoku: Yes, I took inspiration from your changes to the Julia client while working on trying to upgrade an ancient Django project from version 1.9.

I was having some trouble getting the evaluation of classes and decorated classes/functions. In anger, I just did the simplest thing to keep me from confusing myself.

russtoku:
Thanks for making Conjure and making all of this possible!

PR #634, opened 2024-12-24
Refactor log.fnl to remove aniseed nvim module

saccarosium:
Hi,
this PR refactors log.fnl to use neovim builtin api instead of using aniseed nvim module.

### DIFF
--- a/fnl/conjure/log.fnl
+++ b/fnl/conjure/log.fnl
@@ -1,6 +1,5 @@
 (local {: autoload} (require :nfnl.module))
 (local a (autoload :conjure.aniseed.core))
-(local nvim (autoload :conjure.aniseed.nvim))
 (local str (autoload :conjure.aniseed.string))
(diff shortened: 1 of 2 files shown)


saccarosium:
This needs a bit more cooking.

Olical:
Fair enough! Cook away and have a nice new year. And ask away if you have questions about issues you saw, I was going to look at merging this today or tomorrow before it was closed.

saccarosium:
Hey, have nice new year too! I've had some regressions that I didn't notice at first so I need to make look into that further.

I was thinking that the main thing to refactor, at least for my point of view, is a lot of the aniseed cruft and try to use as much as possible the built in APIs. This should make things more appreciable and clear to maintain and to read to newcommers. I think a good goal should be to reduce the learning curve for new contributors and hackers of the code to be able to just use their existing neovim API knowledge.

PR #640, opened 2025-01-15
Refactor treesitter.fnl to remove aniseed nvim module

saccarosium:
Hi,
this PR refactors treesitter.fnl to use neovim builtin api instead of using aniseed nvim module.

### DIFF
--- a/fnl/conjure/tree-sitter.fnl
+++ b/fnl/conjure/tree-sitter.fnl
@@ -1,6 +1,5 @@
 (local {: autoload} (require :nfnl.module))
 (local a (autoload :nfnl.core))
-(local nvim (autoload :conjure.aniseed.nvim))
 (local client (autoload :conjure.client))
 (local config (autoload :conjure.config))
 (local text (autoload :conjure.text))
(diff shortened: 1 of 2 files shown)


Olical:
Nice, I've also just swapped the implementation for starts-with and ends-with out so they just call the built in vim function now. Tests still pass which is reassuring! I was worried about how they would handle `nil` etc. If the `vim.*` ones do NOT handle `nil` well, they might error in unexpected ways. the `conjure.text` functions do handle `nils` which is more Clojure-y and how I like to keep the APIs ideally.

Olical:
Oh, actually now the HUD isn't appearing when I open a file 😬 investigating

saccarosium:
>  I was worried about how they would handle nil etc. If the vim.* ones do NOT handle nil well, they might error in unexpected ways. the conjure.text functions do handle nils which is more Clojure-y and how I like to keep the APIs ideally.

Oh makes sense. By testing the `vim.*` ones do not handle nils and they throw an exeption. So it might be a better idea to use just use `text` for that.

Olical:
Oh no it's okay, I've rolled back commits and I still see the issue. So it seems unrelated... but also new? 🤔

Olical:
Seems like `CursorMoved` is firing when I open a fresh buffer which triggers `log.close-hud-passive` early. This never used to be the case so I suspect the behaviour changed between Neovim versions.

Olical:
https://github.com/Olical/conjure/issues/641 tracking here

PR #649, opened 2025-02-14
Improve selection of forms for Rust client

russtoku:
The addition of a `form-node?` function to the Rust client should make it more convenient to put your cursor on a line and evaluate statements or expressions.

The `dev/rust/sandbox.rs` has more comments to clarify how to evaluating things. I've included a link to the [Evcxr common usage information](https://github.com/evcxr/evcxr/blob/main/COMMON.md) document to help those not very familiar with Rust and `evcxr`.

### DIFF
--- a/dev/rust/sandbox.rs
+++ b/dev/rust/sandbox.rs
@@ -1,11 +1,16 @@
-// evaluate this root form
+// Examples from "Evcxr common usage information".
+//   - https://github.com/evcxr/evcxr/blob/main/COMMON.md
+
+// With the cursor on the word, pub, evaluate form works.
 pub struct User {
     username: String
 }
+
 let user = User { username: String::from("John Doe") };
 println!("{}", user.username);
 
-// select and evaluate this:
+// The REPL will accept commands that are not legal Rust statements.
+// Select and evaluate this to load the rand crate from crates.io:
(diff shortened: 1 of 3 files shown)


russtoku:
Sorry but further comparisons of this mod with the current Rust client make me question the value of this mod. It seems like the only improvement is being able to put your cursor on the keyword, `pub`, and doing a `<localleader>ee` versus `<localleader>er`. Here's the statement from `dev/rust/sandbox.rs`:
```rust
pub struct User {
    username: String
}
```
In other words, the `struct` with the visibility modifier, `pub`, is treated as complete syntactic form versus when evaluating the current form as opposed to being an *outer form* and an *inner form*.

I'm waffling on the mod but I think the changes to `dev/rust/sandbox.rs` are a keeper.

Olical:
I still think that's a valid change, evaluating `pub` on it's own probably doesn't make sense, so having Conjure walk the tree a little more to find something worth evaluating sounds good to me.

Olical:
I gave it a go and it works fine for me. I say we ship this and if power Rust users have Opinions on what should or should not be part of a form then we can take that into account later. This is a good start though and I think most languages (all?) benefit from having some level of filtering of the tree for forms. Just using the node under the cursor is rarely what the user actually wants.

Issue #661, opened 2025-04-28
ConjureSchool script is broken because it tries to source a .vim that is now a .lua


Olical:
Easy fix, just don't have time right now because I'm at work. If anyone wants to fix this, should be a simple one 😄

PR #662, opened 2025-04-28
fix(school): update script to call conjure.lua

practicalli-johnny:
Update Conjure school script to call conjure.lua (was conjure.vim)

resolve: #661


Olical:
Sweet thanks!

PR #666, opened 2025-05-08
Fix unknown-module-field for vim.loop.constants.SIGINT

russtoku:
This is a minor fix to quiet the diagnostics from the `fennel-ls` language server when working with the `stdio`-based Conjure clients.

Just tidying this a bit so feel free to reject this PR.

### DIFF
--- a/fnl/conjure/client/hy/stdio.fnl
+++ b/fnl/conjure/client/hy/stdio.fnl
@@ -158,7 +158,7 @@
   (with-repl-or-warn
     (fn [repl]
       (log.append [(.. comment-prefix " Sending interrupt signal.")] {:break? true})
-      (repl.send-signal vim.loop.constants.SIGINT))))
+      (repl.send-signal :sigint))))
 
 (fn on-filetype []
   (mapping.buf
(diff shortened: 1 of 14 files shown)


Olical:
I need to check but I'm surprised this works. `vim.loop.constants.SIGINT` actually returns `2`, and the signature for `uv_process_kill` (which we call) takes a `int signum` as it's last argument. I worry that the string is being cast to a `1` or something like that so it's accidentally working but through some weird boolean-y casting.

If this is just a linting thing, maybe we can find another way to teach the language server. Or maybe I set up our own map of codes that mirror the vim constants. It is a little wordy to access it at the moment.

Olical:
Nevermind, I found it in the Neovim help text.

```
uv.process_kill({process}, {signame})                         *uv.process_kill()*

                > method form `process:kill(signame)`

                Parameters:
                - `process`: `uv_process_t userdata`
                - `signame`: `string` or `integer` or `nil` (default: `sigterm`)
```

So it does accept strings and there's a mapping of those strings to their signals in the same help text. Maybe you found this already and that's why you did this, just sharing that I found the proof too 😄 thanks for simplifying this, it does read better.

russtoku:
Apologies, my description only stated a fix for an error message. I should have included some explanation about `vim.loop` being deprecated in Neovim 0.10 in favor of `uv.loop` and that the code could be clearer with just the signal name instead of the constant. Although, sometimes constants are preferred to values.

Olical:
Ah okay, a double improvement then!

PR #667, opened 2025-05-14
Make it possible to turn off clojure.test failure pretty printing

philomates:
In some clojure codebases/test setups it is useful to setup custom clojure.test reporting methods, like:

```clojure
(defmethod clojure.test/report :fail [m]
  ;; whatever custom reporting code
  )
```

But since https://github.com/Olical/conjure/pull/569, Conjure might override those, depending on if the nrepl connection is established before or after the codebase's custom clojure.test reporting methods are registered.
And you get conjure's version even in the presence of your own override:

```clojure
(get (methods t/report) :fail)
;; => #function[conjure.internal/eval127846/fn--127847]
```

Thus I figured it would be nice to be able to turn off conjure's registration of the `clojure.test/report` via a config flag.

This PR introduces a new config `:clojure :nrepl :test :pretty_print_test_failures` config field that defaults to true but can be set to false if needed

Somehow even with these changes and setting the new flag to false (`(set nvim.g.conjure#client#clojure#nrepl#test#pretty_print_test_failures false)`) I still get the pretty-printing behavior. Maybe I misunderstood how to go about this.

### DIFF
--- a/doc/conjure-client-clojure-nrepl.txt
+++ b/doc/conjure-client-clojure-nrepl.txt
@@ -313,6 +313,12 @@ All configuration can be set as described in |conjure-configuration|.
             current root form is a test that should be ran.
             Default: `["deftest"]`
 
+                      *g:conjure#client#clojure#nrepl#test#pretty_print_test_failures*
+`g:conjure#client#clojure#nrepl#test#pretty_print_test_failures
+            Pretty print a diff of the test's expected and
+            actual data-structures.
+            Default: `true`
+
                                   *g:conjure#client#clojure#nrepl#test#runner*
 `g:conjure#client#clojure#nrepl#test#runner`
             Test runner to use for the various test mappings. The following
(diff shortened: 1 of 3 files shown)


Olical:
Thanks, this looks good to me! I'm trying it out locally and _also_ struggling to get the config to override, very odd! I'm wondering if there's a bug somewhere in the config system around falsy values or something. Investigating now.

Olical:
Ah so the issue I was having was that `.nvim.lua` wasn't being loaded for me because I didn't have `exrc` set. I must have forgot to set it in my latest dotfiles refresh!

And I think your issue was that the changes you made to the Fennel were never compiled to `.lua` so I just had to go `:w` in those files with nfnl installed 😄 all good now! Merging!

philomates:
Great! Thanks so much for taking the time to try the patch and investigate, really appreciate it

PR #670, opened 2025-05-22
Rewrite bencode using mutable string buffer (fix #669)

### LINKED ISSUE #669: Bencode decode performance issues
Running into issues with the bencode decoding when decoding a large message spread across many chunks. (This does not appear to be an issue with any other repl clients I've tried, such as Elin, Calva, Cider, etc.)

Specifically seeing these large messages from the `track-state` middleware, which causes a several minute Neovim freeze with CPU pinned at 100% when loading in a large library NS. ((use 'com.rpl.rama) [#rama](https://clojurians.slack.com/archives/C05N2M7R6DB))
It seems to return metadata for ~600 variables, which takes ~77 seconds to be fully received and then another couple minutes to fully process.

Currently [decode-all](https://github.com/Olical/conjure/blob/1e656c64494f0b0c51baa40ea3c61f2b608e15b2/fnl/conjure/remote/transport/bencode/init.fnl#L8) is being called for every single chunk which ends up allocating 2 new strings for the entire accumulated chunk so far until it can finally decode a full message.
I think it may be possible to make the bencode implementation accept a buffer table instead of a string, in which case each chunk can efficiently be appended instead of generating new strings every time. (Since LuaJIT doesn't support efficient mutable string operations)
Perhaps a more incremental bencode decoding algorithm that didn't rely on receiving the full message first would be warranted instead. :thinking_face:

Additional context in Slack: https://clojurians.slack.com/archives/CK143P6D7/p1745444129329829

garrett-hopper:
I think this still needs a bit of work to avoid the `(lua "return ...")`s and a few other weird things from my quick and dirty lua translation.

This also uses LuaJit specific [String Buffer Library Extension](https://luajit.org/ext_buffer.html) which may not be portable to all builds of Nvim. (I have no idea how common it is to actually have Nvim not be built with LuaJit these days. 🤔 Further research needed on that front.)

See #669 for the performance issues being addressed by this.


russtoku:
I ran the tests for bencode by hand and they passed.

`:PlenaryBustedDirectory lua/conjure-spec/remote/transport/bencode_spec.lua`

```
Testing:   ...lua/conjure-spec/remote/transport/bencode_spec.lua

Success || conjure.remote.transport.bencode basic data starts empty
Success || conjure.remote.transport.bencode basic a single bencoded value
Success || conjure.remote.transport.bencode basic data is empty after a decode
Success || conjure.remote.transport.bencode multiple-values data starts empty
Success || conjure.remote.transport.bencode multiple-values two bencoded values
Success || conjure.remote.transport.bencode multiple-values data is empty after a decode
Success || conjure.remote.transport.bencode partial-values data starts empty
Success || conjure.remote.transport.bencode partial-values first value
Success || conjure.remote.transport.bencode partial-values after first, data contains partial data-b
Success || conjure.remote.transport.bencode partial-values second value after rest of data
Success || conjure.remote.transport.bencode partial-values data is empty after a decode

Success:   11
Failed :   0
Errors :   0
```

There should probably be a new test for the handling of many chunks of encoded data.

I haven't looked at the test code so I'm not quite sure how to create such a test.

Olical:
Neovim always uses LuaJIT which makes our lives a little easier 😄 I think if the Lua implementation was switched out there would be riots in the street. It's a safe bet to use LuaJIT specific functions, don't worry about that.

Olical:
This is a (and I can't stress this enough) _really_ cool change. The code is a bit of a black box to me and a little magical right now, but that's not a bad thing if the tests pass and it feels okay to use. I grabbed the original implementation "off the shelf" and didn't think much of it but have also noticed poor performance at times.

This is an incredible improvement, thank you so much. I'll give it a go, maybe add some more tests as well as try to do a little refactoring of my own and see where I get to.

I might also add a little benchmarking system for this and other parts of Conjure so we can have some hard data around changes like this in a reproducible way.

garrett-hopper:
I think depending on LuaJit is probably fine as well. My hesitation was based on [:h lua-compat](https://neovim.io/doc/user/lua.html#lua-compat)

> Lua 5.1 is the permanent interface for Nvim Lua. Plugins should target Lua 5.1 as specified in |luaref|; later versions (which are essentially different, incompatible, dialects) are not supported. This includes extensions such as `goto` that some Lua 5.1 interpreters like LuaJIT may support.

And the subsequent [:h lua-jit](https://neovim.io/doc/user/lua.html#lua-jit)

>While Nvim officially only requires Lua 5.1 support, it should be built with LuaJIT or a compatible fork on supported platforms for performance reasons. LuaJIT also comes with useful extensions such as `ffi`, |lua-profile|, and enhanced standard library functions; these cannot be assumed to be available, and Lua code in |init.lua| or plugins should check the `jit` global variable before using them: 
```lua
  if jit then
    -- code for luajit
  else
    -- code for plain lua 5.1
  end
```

Perhaps worth a `if jit` check? Though I'm sure many other plugins depend on it without doing anything to handle its absence. We could always keep a compatible version around for when it's not, but 🤷.

---

Sorry about the black box; was trying to finish this late at night so I could close the loop and stop thinking about it. 😆 Going to take another look at it today to add some additional comments and try to make it more lispy without the return escape hatches.

@Olical, let me know if you're actively looking at this or focusing on the testing/benchmarking; don't want to step on your toes.

Very interested in benchmarks. I'm not sure the performance different will be noticeable for anything but projects with large dependencies on the classpath. (Even then, I'm still wondering if the `track-state` middleware should really be sending down private variable metadata which is what bloats the message across many bencode stream chunks, but that's a separate issue.)

Olical:
Ah okay I didn't realise it could be swapped out. Please feel free to make the assumption of LuaJIT though, I like targeting JIT + 5.1 as a known interpreter. I'd rather not add complexity to the system to support an interpreter that may never run any of this. We'll cross that bridge if we ever come to it 😃 

And go wild on this PR, I'm away at the moment so just looking in here and there. Working on a little benchmarking suite on the `main` branch at the moment and will run it against your code once it's ready.

Thanks for all of the effort! It's appreciated! I've also enabled the test suite GitHub action against PRs now so you should get the test suite running on each push now I hope.

Olical:
Benchmarking results! You can get this by rebasing on the main branch and running `mise benchmark` although I appreciate mise isn't super popular, I just love it. I can try to find a way to make it easy for everyone to execute.

## Original on the `main` branch

```
Iterations: 1000
# bencode
## simple encode and decode [0.017ms]
## big encode, chunked decode [58.639ms]
```

## Your branch

```
Iterations: 1000
# bencode
## simple encode and decode [0.014ms]
## big encode, chunked decode [3.497ms]
```

A great improvement! This is an average of 1000 iterations, the code I'm executing for the big encode and chunked decode involves encoding 500 small-ish objects then chunking that string into 512 character chunks and passing each to the parser.

If I don't chunk it, the speed is kind of comparable, but the speed is MUCH better when given chunked large input. So this is a great move. I hope this tooling can help you develop this!

I've been running it with `mise watch benchmark`. I highly recommend [mise](mise.jdx.dev) but will give you a simple script to execute instead if you prefer not to go through the hassle of a new tool.

garrett-hopper:
@Olical, I think I've finally got this to a place I'm happy with. Thank you for writing those benchmarks, they proved super helpful for validating this. I was able to find several edge cases thanks to them. Took me forever to figure out I was seeing flaky tests due to my encode function not actually sorting keys. 😆

Let me know what you think. Feel free to make or request any changes you see fit to this branch.

russtoku:
I'm seeing some test failures for `bencode` when I run `scripts/test` on this PR's code. With the main branch's code, I don't get any failure for `bencode`. 

```console
Testing: 	.../conjure/lua/conjure-spec/remote/transport/bencode_spec.lua	
Fail	||	conjure.remote.transport.bencode basic data starts empty	
            Expected objects to be equal.
            Passed in:
            (string) ''
            Expected:
            (nil)

            stack traceback:
            	...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:14: in function <...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:8>
            	
Success	||	conjure.remote.transport.bencode basic a single bencoded value	
Fail	||	conjure.remote.transport.bencode basic data is empty after a decode	
            Expected objects to be equal.
            Passed in:
            (string) ''
            Expected:
            (nil)

            stack traceback:
            	...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:24: in function <...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:7>
            	
Fail	||	conjure.remote.transport.bencode multiple-values data starts empty	
            Expected objects to be equal.
            Passed in:
            (string) ''
            Expected:
            (nil)

            stack traceback:
            	...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:32: in function <...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:25>
            	
Success	||	conjure.remote.transport.bencode multiple-values two bencoded values	
Fail	||	conjure.remote.transport.bencode multiple-values data is empty after a decode	
            Expected objects to be equal.
            Passed in:
            (string) ''
            Expected:
            (nil)

            stack traceback:
            	...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:42: in function <...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:7>
            	
Fail	||	conjure.remote.transport.bencode partial-values data starts empty	
            Expected objects to be equal.
            Passed in:
            (string) ''
            Expected:
            (nil)

            stack traceback:
            	...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:51: in function <...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:43>
            	
Success	||	conjure.remote.transport.bencode partial-values first value	
Fail	||	conjure.remote.transport.bencode partial-values after first, data contains partial data-b	
            Expected objects to be equal.
            Passed in:
            (nil)
            Expected:
            (string) 'li1'

            stack traceback:
            	...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:59: in function <...njure/lua/conjure-spec/remote/transport/bencode_spec.lua:43>
            	
Success	||	conjure.remote.transport.bencode partial-values second value after rest of data	
Fail	||	conjure.remote.transport.bencode partial-values data is empty after a decode	
            error in error handling	
	
Success: 	4	
Failed : 	7	
Errors : 	0	
```

garrett-hopper:
Tests fixed and several new ones added. 🙂

russtoku:
`scripts/test` reported all 38 tests for `bencode` succeeded. Nice!

Olical (approved):
This is pretty damn cool [extreme understatement] and should probably be released as an awesome standalone library. If you did I'd happily embed it under Conjure so you can have more freedom to develop it over time and share it with more people.

Thank you so much for this work, I really hope it helps speed things up for you and many many others. This will be especially helpful if I continue down the path of trying to make most if not all clients based on nREPL + bencode in the future. (still just an idea right now https://github.com/Olical/nyREPL - will get to it when my stack of side projects pops up a couple of levels)

garrett-hopper:
Thanks! 🎉

PR #676, opened 2025-06-26
R (stdio) client

perrygeo:
This is a copy-paste job, forked from the scheme stdio client. I've been using it internally for a few months, figured I'd share here to see if it's worth merging upstream.

What else is required to make this an official conjure client?

image

I'm not entirely happy with the output - R likes to respond with a variable number of "+" symbols or nothing, both of which would typically print "nil" to the conjure log using Clojure. It also prefixes the output with a number (eg "[1] 7" instead of "7"). Any ideas on how to fix these quirks? They're not deal breakers for me (it's useful as is) but it might prevent you from merging this to main?


russtoku:
Oh, nice!

Sorry I won't try it because I don't use R but thanks for investing your time in it.

Olical:
Oh very interesting! I've also never used R but I'll take a look at this PR and see what advice I can offer.

Olical:
This change will be required to get the client to load when entering an R filetype too:

```diff
diff --git a/fnl/conjure/config.fnl b/fnl/conjure/config.fnl
index 2584e311..7a1fd482 100644
--- a/fnl/conjure/config.fnl
+++ b/fnl/conjure/config.fnl
@@ -54,7 +54,7 @@
    :path_subs nil
    :client_on_load true

-   :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust :sql]
+   :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust :sql :r]
    :filetype {:clojure :conjure.client.clojure.nrepl
               :fennel :conjure.client.fennel.nfnl
               :janet :conjure.client.janet.netrepl
```

We'll need to create `doc/conjure-client-r-stdio.txt` based on the Scheme client help text. There's also wiki pages for each client just to help people get started with each language, those are linked to from the README.

I can always make these changes and create these help files though, and then once we have the core stuff working and documented I'll merge and apply some modernisation to how the module is written. Not your fault there, most of the client are "outdated" and I am slowly upgrading them to avoid using Aniseed at all and to use the `nfnl.module` `define` syntax documented here https://oli.me.uk/Guides/Reloadable+Fennel+in+Neovim

Overall though, this is really cool and it works well. I think those number prefixes might just have to stay since they're part of the core REPL output and presumably have some use? The empty output / lines of `+` symbols maybe needs more understanding and cleaning although it's not a dealbreaker for me either.

perrygeo:
Good catch, I set the R filetype in my personal vim config and forgot about this.

I'll work on the docs and a wiki page next week (traveling this weekend). 

The output quirks are part of the R REPL experience I suppose. Probably not worth trying to modify it - R users might even expect it.

Olical:
Yep exactly, some of those things might be important in some cases. We could spend 2 hours writing code to strip things out then realise some people need some of it, then you end up with a bunch of options, config and edge cases. Sometimes it's better to just let everything flow out so you're sort of close to a raw REPL, but in a regular buffer.

If you don't have the time to do any help text copy/paste/editing etc then just let me know, we can get this merged anyway and I can jump on the documenting after it's merged.

Olical:
Done! Rebased on `main`, conflicts resolved and merged. Thank you very much! This is really awesome. We got PHP and R in one day, that's another few potential users 😄 

Soon the answer to "what does Conjure support?" will just be "yes".

PR #678, opened 2025-06-28
add global config for raw out

ThatAmuzak:
Add a global configuration option to get raw outputs instead of prepending comment strings and (out)/(err) to outputs.
This is meant to elevate the language specific options from `clojure` and `hy` to be applicable to languages such as `python`, `rust`, `scheme`, and `lua`.
This also overwrites the existing language specific options from `clojure` and `hy`, but leaves the `clojure` test raw output option as is (at least for now, wasn't sure if that should be in the scope of this option. Maybe log -> eval -> raw_out and log -> test -> raw_out can be two separate options and then each language can choose to implement it based on requirements)

Not fully sure where the equivalent comment string outputs are for the languages that haven't been modified here, need help from the maintainers/others interested in such a feature. It's marked as draft for the time being to address these issues.

---

Note: **_I have absolutely no clue what I'm doing_**. I know little lua, and even lesser lisp (just did a quick primer before starting this PR 😅).
From the list of languages that Conjure supports, I only understand python and lua (by virtue of neovim).
There's probably a bunch of cases that I've missed, and I wasn't entirely sure in the languages that haven't been modified so far what the out equivalents were.
I did setup nfnl to compile the files to Lua to see if there were any errors and saw none, but I could be missing something there as well.
Looking for some help in completing this PR!


Olical:
Ohh this is looking good! Well done on working out so much so quickly, you've done a great job here! This is a hard thing to do because it requires fiddling in each of the clients in specific and different ways. This is a great start and I'll try to help you get this across the line.

I am on a (short) train right now so I don't have long but my initial thoughts are:

 1. This branch of logic will repeat in lots of places and will be kind of non obvious because it'll look different in every client.
 2. Clients print stdout/stderr (which is what we care about for this option if I'm not mistaken) in subtly different ways.
 3. This technically introduces a breaking change if we remove the existing options used in clients.

So, to address these: I propose we add a function to `conjure.log` that unifies how stdout/stderr are printed, that function will know how to format these strings (`comment-prefix (kind of output) message`) which will be the only thing that needs to look at the `raw_out` option. That way the branching logic _and_ formatting of the output is contained in one place.

We'd have to pass an optional table to the function that can override the global `raw_out` with the legacy client local options until they are removed. I'll probably look at adding a warning to users that use the client local options to help them migrate to the global one. If users then still need raw_out in Python for example but in nothing else, they can use an autocmd to set a `b:` (buffer local) option which won't affect other filetypes.

If you're not sure how to proceed with this or only want to attempt some of it, that's totally fine. I can help you finish it off and add some commits on top soon. But I'll give you some time just in case you want to have a go yourself.

Please ask away if you have any questions, I'll do what I can to help! And I wouldn't worry about getting this working in every language just yet, if we can get some good reusable functions that check this option then we can migrate the rest when we're happy.

The best changes in Conjure are the ones that don't touch each client file but instead operate above them, so one change helps ALL clients, so if we can pivot this slightly from "changes mostly happen inside every client" to "changes happen above them with a small amount of integration code in each one so they share a codepath" I think that'll be higher leverage over time. Gives us another single function that touches all clients, fulcrums, yay!

ThatAmuzak:
I'm trying to understand what this would look like, and here's what I have in mind (to account for the different behaviours per file type as well)

User defines something like: 
```lua
init = function()
-- other configuration options

  vim.g["conjure#log#std_out"] = {
    raw_out = true --(false by default)
    file_types = {"python"} --(all languages by default)
  }
  vim.g["conjure#log#std_warn"] = {
    -- same options and defaults as above
  }
  vim.g["conjure#log#std_err"] = {
    -- same options and defaults as above
  }

end
```

in `fnl/conjure/log.fnl`, we have functions that would look as follows:

```fnl
(fn std-out [msg prefix language ?legacy-override]
  ;; perform the legacy check, else use config vals from above
  ;; then write to repl
  )
;; similar functions for warn and err
```

And then each language hooks into the relevant log function as necessary, passing in the legacy override if required from clojure or hy.

The reason I'm thinking along the lines of separate config options for std_out, warn and err is for potentially different preferences for different logs, like I want warnings commented, but I would want normal logs and errors to not be commented out. 
We could wrap it in a single larger table, just not sure what a good name would be (maybe just std? And then each sub table is like out, warn, err, with the next nest being raw, and file types to affect).

The other thing is the behaviour for file types, which I thought would be interesting.
For instance, if I said raw out true for file types python, then I would intuitively expect the output for clojure to be commented out unless added to that file type list, and vice versa. So the log function would utilize the language data that it's being passed in from, and perform that check. Not sure if there is a fennel specific way of seeing which class the function is being called from?

Let me know if this sounds good, and I'll start work on it!

russtoku:
With respect to:

> I propose we add a function to conjure.log that unifies how stdout/stderr are printed, that function will know how to format these strings (comment-prefix (kind of output) message) which will be the only thing that needs to look at the raw_out option. That way the branching logic and formatting of the output is contained in one place.

I'm in favor of leaving formatting decisions to the client so that clients just call a log function with the message that they want to appear in their log buffer. Having *comment-prefix*, *kind of output*, and *raw_output* assumes that there might not be other ways to munge the output lines.

To assist clients, we could create a helper function, `format-output`, that would take *(comment-prefix (kind of output) message)* and return the message lines commented out. `format-output` could also centralize the *to raw or not to raw*  decision there. Optionally, clients can provide their own formatters and call that before calling the log function. Also, to ease the transition, we can add a plain log function to be used together with `format-output` for a "unified" function call pattern. Hopefully, this strikes a balance between

> "changes mostly happen inside every client" to "changes happen above them with a small amount of integration code in each one so they share a codepath"

How about `source of output` instead of `kind of output`?

With respect to:

> So the log function would utilize the language data that it's being passed in from, and perform that check.

I'm not in favor of giving a log function used by many clients the responsibility of knowing about different client languages. Language-specific things should be handled by clients and not included in the log function. This *de-complects* the language-specific concerns from the basic log function of puttting output messages into the log buffer for a client. If you're not already familiar with the term, *complect*, Rich Hickey explains it in https://www.youtube.com/watch?v=LKtk3HCgTa8&t=1944s.

russtoku:
I don't think these items should be in`in fnl/conjure/config.fnl` because it exposes client-level decision at the higher plugin level.

```fennel
 :log
   {:wrap false
     ...
    :std_out {:raw false
              :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust :sql]}
    :std_warn {:raw false
              :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust :sql]}
    :std_err {:raw false
              :filetypes [:clojure :fennel :janet :hy :julia :racket :scheme :lua :lisp :python :rust :sql]}}
     ...
```
Instead, I think they should be client-specific like:

```fennel
(config.merge
  {:client
   {:clojure
    {:nrepl
     {:std_out {:raw false}
      :std_warn {:raw false}
      :std_err (:raw false}
    ...
```

This makes it easier for users to understand what is configurable for a client they want to use. The help doc for a client might read something like:

```text
g:conjure#client#clojure#nrepl#std_out#raw
            Whether or not to decorate the std_out messages or no.
            Default: false

g:conjure#client#clojure#nrepl#std_warn#raw
            Whether or not to decorate the std_warn messages.
            Default: false

g:conjure#client#clojure#nrepl#std_err#raw
            Whether or not to decorate the std_err messages.
            Default: false
```
---

I'm not really clear about what you meant by`std_out`, `std_warn`, and `std_err` when you said:

> I'm thinking along the lines of separate config options for std_out, warn and err is for potentially different preferences for different logs, like I want warnings commented, but I would want normal logs and errors to not be commented out.

So, I'm guessing that your terms are referring to the actual content of messages rather the source channels.

Typically, *stdout* and *stderr* are used to refer to the `stdout` and the `stderr` streams from a process. The [Overview of nREPL](https://nrepl.org/nrepl/design/overview.html) says that responses from an nREPL server will have keys mapping  `:out` and `:err` to the messages from the *stdout* and *stderr* channels. This aligns with Clojure's [\*out\*](https://clojuredocs.org/clojure.core/*out*) and [\*err\*](https://clojuredocs.org/clojure.core/*err*).

Also be aware that the *stdio* clients (Python, etc.) have *stdout* and *stderr* streams while the Guile *socket* client has the *stdout* and *stderr* streams merged into one, the *stdout*.

ThatAmuzak:
Hmm, I was initially looking for a global level config for getting raw out across clients, from purely a convenience to configure standpoint. But maybe the easy solution isn't the simple one here 😁

And yes, my understanding was the formatting was based on the content of the message, but given that these are the sources available, and some clients only have specific streams, maybe it's better to do it per client. I'll move to that implementation, and see which clients have which streams available, and allow options on the basis of that

ThatAmuzak:
After a bit more digging, I'm starting to realize this isn't that trivial a task if we want to have global options for raw output across multiple clients, at least for a complete novice like me (without making a mess of how data flows between different modules). There's already a function for [prefixing results](https://github.com/Olical/conjure/blob/main/fnl/conjure/text.fnl#L25), but that one has logic to account for simple out, which seems to be more of a clojure client specific option from what I understand (could be wrong about this). 

I'll probably revisit this when I have some more time on my hands to learn fennel and conjure better, or leave it for someone more competent to take up. For now I'll just make a hacky fix on my own fork, and see how that plays out for a while. Thanks for helping me out in attempting to approach the issue!

PR #681, opened 2025-07-03
feat: Add php (psysh) client

atomicptr:
An attempt at adding a PHP client (using the [psysh](https://github.com/bobthecow/psysh) repl)

https://github.com/user-attachments/assets/d18de47e-0534-4865-98fe-23360a437f39


Olical:
Rebased on `main`, works for me! Although I'd love more structure to the output so we don't put a comment in front of all return values, this is an excellent start. I'm hoping I'll be able to add all that structure in https://github.com/Olical/nyREPL when I get to it, so the client can simplify and just connect that that wrapper server which hides `psysh` away from Conjure entirely.

Until then, I'm very happy with this start, thank you!

PR #682, opened 2025-07-03
Use nfnl define syntax for Python client to enable hot re-loading

russtoku:
Followed the changes made to the Guile client to update the Python client.

### DIFF
--- a/fnl/conjure/client/python/stdio.fnl
+++ b/fnl/conjure/client/python/stdio.fnl
@@ -1,4 +1,4 @@
-(local {: autoload} (require :conjure.nfnl.module))
+(local {: autoload : define} (require :conjure.nfnl.module))
 (local a (autoload :conjure.aniseed.core))
 (local str (autoload :conjure.aniseed.string))
 (local stdio (autoload :conjure.remote.stdio))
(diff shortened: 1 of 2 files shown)


fnl/conjure/client/python/stdio.fnl
```diff
@@ -256,61 +257,43 @@
...
   (mapping.buf
     :PythonStop (cfg [:mapping :stop])
-    stop
+    M.stop
     {:desc "Stop the Python REPL"})
 
   (mapping.buf
     :PythonInterrupt (cfg [:mapping :interrupt])
-    interrupt
+    M.interrupt
```
Olical, line 296: Ah yeah, here! This is what my email was mentioning, if this was `#(M.interrupt)` then it'll be hot reload friendly, as it stands this points the mapping at v1 of this function and even when you reload it the old version is still used by the mapping.

Olical (approved):
This looks great! The only change I'd make is a layer of indirection in the callbacks to make them hot reloadable too.

Olical:
Rebased and merged with the changes I suggested to the mapping callbacks, thanks, Russ!

PR #685, opened 2025-07-08
Conjure JavaScript Support

Stansom:
Hi, I've added JavaScript support for Conjure. Because I use JS quite a lot, REPL support really helps me move faster and make fewer mistakes, providing a pretty similar experience to Clojure.

However, the Node.js REPL doesn't allow "dynamic rebinding" of vars. Functions declared using the "const fnName = (...args) =>" form are not re-declarable. I've introduced some transformations from the "const" form into the standard JS function form, so you don't have to constantly keep that Node.js quirk in mind. Also, the Node.js REPL doesn't support "import". My solution is to change "import" into the "const importName = require(...)" form, so you can use it easily.

JavaScript support helps me make quick sketches and experiments, and test my code immediately, just like I do in Clojure. I hope someone else finds it helpful too.


Stansom:
Hi! Sandbox has been added.

russtoku:
The `sandbox.js` file looks great! Even with async stuff.

Almost all of the lines in `sandbox.js` can be evaluated with `<localleader>ee` except:
- `import` statements
- `class` definition

The `import` statements and `class` definition have to be selected and evaluated with `<localleader>E`.

Thanks for creating a very nice client! Works smoothly from what I can tell (I'm an infrequent user of JavaScript).

russtoku:
I have some suggestions:

- To make it so that `<localleader>ee` works with `import` statements and `class` definintions, add the following line to the `form-node?` function:
    ```fennel
    (= :import_statement (node:type)) (= :class_declaration (node:type))
    ``` 
    It can be added after the line that reads:
    ```fennel
    (= :try_statement (node:type)) (= :expression_statement (node:type))
    ```
- Format the Fennel code like the other clients. This is for consistency. While there is a Fennel [style guide](https://fennel-lang.org/style), it may be easier to follow the other clients' formatting.
- Try to handle messages sent out via stderr. This code will send a string out via stderr but nothing appears in the log.
    ```fennel
    process.stderr.write("error! some error occurred\n");
    ```
- Use `nfnl`'s `define` to make the client reloadable. This makes it easier to develop and maintain the client as you won't have to restart Neovim to see the effects of changes to the Fennel code. For an example, see `fnl/conjure/client/guile/socket.fnl`.

Stansom:
- I usually use <localleader>er and didn't even notice that import and class don't work with <localleader>ee. Fixed.
- Formatted the code to match other clients style.
- Used nfnl's define, but I still don't understand how it works. 😅
- Regarding processing messages in stderr, unfortunately, I couldn't figure out where I could intercept them to output to the REPL.

russtoku:
@Olical has a guide for [Reloadable Fennel in Neovim](https://oli.me.uk/Guides/Reloadable+Fennel+in+Neovim).

 I hope that I got this correct. I believe the way it works is the `define`macro in `nfnl` sets things up so that using a `#(M.func)` form will call the `M.func` in the most recently compiled version of the `M` module rather than the version of the module that existed when Neovim was started up for the session. In other words, a bit of indirection magic calls the current compiled version.

I found that you first compile the module's file or the entire buffer (of the module's file if you're editing it). After that, you can switch back to a code buffer and see the new behavior of the client when you evaluate pieces of code.

I find that another benefit of using `nfnl`'s `define` macro is that it's easier to know which functions and variables are publicly visible when you're reading or working with the code. Otherwise, you'd have to repeatedly jump to the bottom of the file to see which ones are included in the table that's returned when the Lua file is required.

russtoku:
Test your latest changes and they work.

I think you don't have to worry too much about processing messages via stderr. That can come later. What you have now works well.

I'll dig into my old notes and see if I can come up with some suggestions.

Stansom:
@russtoku Thank you for your interest and suggestions! I will try to tinker with stderr later.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -0,0 +1,271 @@
+(local {: autoload : define} (require :conjure.nfnl.module))
+(local a (autoload :nfnl.core))
+(local str (autoload :nfnl.string))
```
Olical, line 3: These need to be `conjure.nfnl.string` etc. This line will work if you have nfnl installed as well but won't work for users that ONLY have Conjure installed. The `conjure.` prefix means we'll use the copy bundled into this repository.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -0,0 +1,271 @@
...
+(fn is-arrow-fn? [s]
+  (if (not= :string (type s)) false)
+  (let [ts (s:match "^%s*(.-)%s*$")
+        expr (or (ts:match "=%s*(.*)") ts)
+        parens "^%s*%b()%s*=>"
+        ident "^%s*[%a_$][%w_$]*%s*=>"]
+    (if (not (or (ts:find "=> ") (ts:find "%f[%w]function%f[%W]")))
+        false)
+    (if (expr:match parens) true)
+    (if (expr:match ident) true)
```
Olical, line 98: This won't work as you intend, there is no such thing as an early return in Fennel (or any lisps really (exclusions may apply)). Although there are some hacks that allow you to insert raw Lua into the compiled output which can let you insert a `return`, it's a huge hack and not used.

Instead everything must be nested in expressions, it takes a little while to get out of the habit of early returns and instead into the habit of nested or chained conditional expressions.

This function could be written something like this (could be wrong, I'm writing this on a machine with no coding tools so I can't check it just yet):

```fennel
(fn is-arrow-fn? [s]
  (if (not= :string (type s))
    false
    (let [ts (s:match "^%s*(.-)%s*$")
          expr (or (ts:match "=%s*(.*)") ts)
          parens "^%s*%b()%s*=>"
          ident "^%s*[%a_$][%w_$]*%s*=>"]
      (if
        (not (or (ts:find "=> ") (ts:find "%f[%w]function%f[%W]"))) false
        (expr:match parens) true
        (expr:match ident) true
        false))))
```

The takeaways here are:

 - There's no such thing as an early return in Fennel (or any expression based language afaik? In general anyway, return is for more procedural C like languages which is effectively a `goto` (which we all avoid!) and I find EXTREMELY confusing and the source of many bugs when I end up working with them after years in Clojure)
 - `if` statements in Fennel can have ANY number of branches, it's like an if else if else if else chain in Lua.
 - If you have _one_ branch of an if statement then use `when` - it's the same as writing `(if X (do Y) nil)` but instead it's just `(when X Y)`. A `when` lets you have many expressions inside of it and is designed for a single condition doing multiple things. You don't need that here, but it's good to know about `if` and `when`.
 - Fennel will have more nesting than you're used to. JavaScript tends to promote early returns to avoid nesting (at the cost of having to spot the returns on ANY line to understand the flow of the code), Fennel requires explicit nesting to return things because we can only return from expressions.

I hope this helps!

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -0,0 +1,271 @@
...
+
+(fn prep-code [s]
+  (.. 
+    (->> (str.split (replace-arrows s) "\n")
+         (a.filter #(not= "" $1))
+         (a.map #(-> $1
+                     str.trim 
+                     replace-imports
+                     (replace-require-path (vim.uv.fs_realpath (vim.fn.expand "%:p:h")))))
+         (str.join "\n")) "\n"))
```
Olical, line 119: Another technique for this would be:

```fennel
(-> (str.split (replace-arrows s) "\n")
    (->> (a.filter ...)
         ...)
    (.. "\n"))
```

So you start with a `->` and your "single value" functions that operate on strings and maps and stuff. Then when you split into many values you start using `->>` because functions that operate on multiple things (by convention borrowed from Clojure) always put the "things" in the last argument. When you're done and you join back from `[x, y, z]` to `x` then you are back in `->` land and can operate on a single value again.

This is a common pattern in Clojure and thus in Fennel too since it shares some syntactic DNA.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -0,0 +1,271 @@
...
+             (log.dbg (-> [msg] 
+                          M.unbatch
+                          M.format-msg)
+                      {:join-first? true}))}))))
+
+(fn warning-msg []
+  (a.map #(log.append [$1])
+         ["// WARNING! Node.js REPL limitations require transformations:"
+          "// 1. ES6 'import' statements are converted to 'require(...)' calls."
+          "// 2. Arrow functions ('const fn = () => ...') are converted to 'function fn() ...' declarations to allow re-definition."]))
```
Olical, line 232: Is the issue here not the use of `const` and we just need to transform that to `let`? Or is it specifically related to arrow functions somehow?

Olical (changes requested):
A great start! I have some comments on some Fennel code but I'll help you work through that where required. Thanks for putting the effort into learning so much for this! I know it's a bit of a mountain to climb but you're doing great so far.

fnl/conjure/client/javascript/stdio.fnl
Stansom: Wow! Thank you very much for such a high quality and detailed comment, my mistake is that I didn't fully understand how if works in Fennel, I looked at 'case' thinking it's the same expression(macro?) as 'cond' in Clojure, but I didn't even think it could be an 'if'. The approach in Fennel looks elegant, in Clojure is more understandable for JavaScript guys 😊

fnl/conjure/client/javascript/stdio.fnl
Stansom: yes it looks and reads much better that way, thanks a lot for the explanation.

Olical:
Giving this a go locally and it seems to be working great for the most part. I ran into an issue with `<prefix>ef` on the sandbox file though, I get the following stack trace.

```
E5108: Error executing lua: Vim:Error executing Lua callback: ...s/Olical/conjure/lua/conjure/client/javascript/stdio.lua:110: attempt to concatenate local 'decl' (a nil value)                                                      
stack traceback:                                                                                                                                                                                                                                
        ...s/Olical/conjure/lua/conjure/client/javascript/stdio.lua:110: in function 'replace_arrows'                                                                                                                                           
        ...s/Olical/conjure/lua/conjure/client/javascript/stdio.lua:138: in function 'prep_code'                                                                                                                                                
        ...s/Olical/conjure/lua/conjure/client/javascript/stdio.lua:188: in function <...s/Olical/conjure/lua/conjure/client/javascript/stdio.lua:178>                                                                                          
        [C]: at 0x5eb7df020c30                                                                                                                                                                                                                  
stack traceback:                                                                                                                                                                                                                                
        [C]: at 0x5eb7df020c30
```

Looks like this if statement just falls through to a nil sometimes then when we try to concat it with a string it throws an exception because `(.. nil "foo")` throws. Ways around this include changing the if statement so a string is always in `decl` or maybe using `conjure.nfnl.core` which has a `join` function that works a bit like `..` but is safe with all kinds of values like `nil` and can take an optional separator to put between each value.

Olical:
If I delete the very first line from the sandbox and re-eval the whole file I get the following:

```
// eval (file): /home/olical/repos/Olical/conjure/dev/javascript/sandbox.js
(out) | | | | | | | | | let a = "foo";
(out)     ^
(out) Uncaught SyntaxError: Identifier 'a' has already been declared
```

Which is interesting!

Stansom:
@Olical Hi! Yes it was definitely a bug that I fixed, now <localleader>ef works. 
```javascript 
// eval (file): /home/olical/repos/Olical/conjure/dev/javascript/sandbox.js
(out) | | | | | | | | | let a = "foo";
(out)     ^
(out) Uncaught SyntaxError: Identifier 'a' has already been declared
```

And that one is a result of the first one.

russtoku:
I ran through the `sandbox.js` file and things work as expected.

Stansom:
Hello @russtoku, your suggestions are always on point! Thank you.

gbroques:
I just found out about Conjure a couple days ago, and got it setup to help write and evaluate Lua code for Neovim.

It's amazing!

I also write a fair amount of JavaScript so I'm looking forward to these changes. Thank you @Stansom for initiating it.

It's great to see all the collaboration and support in this PR from @russtoku and @Olical.

Great work guys. 👏

russtoku:
@gbroques, Thanks for the feedback!

Stansom:
@Olical I wanted to thank you for this fantastic plugin. It was the main reason I switched from VSCode to NeoVim. I really wanted that quick feedback loop when working with Clojure, something VSCode's equivalent plugins couldn't provide at the time. When I saw how simple it was to work with the Clojure REPL in NeoVim using Conjure, I was instantly sold. It took some time to get used to, but the benefits have been huge.

However, when working with JavaScript, I still had to go back to VSCode to get anything resembling “REPL-like” features.  I wanted to write JavaScript in the “Clojure way,” with the same tight feedback loop for exploring and experimenting with code. Unfortunately, VSCode plugins couldn’t match the experience, they lacked proper support for imports and top-level asynchronous code.

One day, I decided to look into Conjure’s source code to see if I could implement a client for a JavaScript REPL. I was delighted to find the sources were easy to understand. Your plugin's architecture is incredibly extensible, and the surrounding ecosystem is so well-designed that it allowed me to move quickly and have a great time building the client. The project is well-structured, the automatic compilation from Fennel to Lua is beautiful, and writing Fennel with Conjure is just so cool and fun.  Thank you for creating such a great tool!

Thank you as well, @russtoku . I really appreciate your quick feedback and helpful suggestions. Your involvement made a big difference!

I hope this small contribution helps JavaScript developers write code faster and enjoy the “LISP way” of coding, with instant feedback and plenty of joy.

PR #687, opened 2025-07-14
Prevalidate scheme and Guile before eval

seamusriordan:
Incomplete expressions will leave the Guile and scheme REPLs waiting for continued input effectively breaking it until it is disconnected/reconnected.  To help preserve health of the REPL we can use tree sitter to at least make sure strings to be evaluated at least parse before evaluation.

I was unable to determine a different kind of check to see if the REPL would be responsive or if there was something we can feed it to put it in a known state, but it seems to me like there's no observable difference between waiting for more input or a long-running evaluation.

Consider the code

```
(define x 33
x
```

Evaluating this either as a buffer or as a root form will silently hang the REPL which will not produce further output until the connection is restarted.

This PR places a check around any code to be evaluated through Guile's and scheme's `eval-str` and yields an error message instead of hanging on an unparseable evaluation.

### DIFF
--- a/fnl/conjure-spec/client/guile/socket_spec.fnl
+++ b/fnl/conjure-spec/client/guile/socket_spec.fnl
@@ -3,7 +3,11 @@
 (local guile (require :conjure.client.guile.socket))
 (local config (require :conjure.config))
 (local fake-socket (require :conjure-spec.client.guile.fake-socket))
+(local ts (require :conjure.tree-sitter))
 (require :conjure-spec.assertions)
+
+(tset package.loaded "conjure.remote.socket" fake-socket)
+(tset ts :valid-str? (fn [_ _] true))
 
 (local completion-code-define-match "%(define%* %(%%conjure:get%-guile%-completions")
 
@@ -13,7 +17,6 @@
 
 (describe "conjure.client.guile.socket"
   (fn []
-    (tset package.loaded "conjure.remote.socket" fake-socket)
     (describe "context extraction"
       (fn [] 
         (it "returns nil for hello world"
--- a/fnl/conjure/client/scheme/stdio.fnl
+++ b/fnl/conjure/client/scheme/stdio.fnl
@@ -65,13 +65,15 @@
 (fn eval-str [opts]
   (with-repl-or-warn
     (fn [repl]
-      (repl.send
+      (if (ts.valid-str? :scheme opts.code) 
+       (repl.send
         (.. opts.code "\n")
         (fn [msgs]
           (let [msgs (-> msgs unbatch format-msg)]
             (opts.on-result (a.last msgs))
             (log.append msgs)))
-        {:batch? true}))))
+        {:batch? true})
+       (log.append [(.. comment-prefix "eval error: could not parse form")])))))
 
 (fn eval-file [opts]
   (eval-str (a.assoc opts :code (.. "(load \"" opts.file-path "\")"))))
--- a/fnl/conjure/tree-sitter.fnl
+++ b/fnl/conjure/tree-sitter.fnl
@@ -200,6 +200,14 @@
        vim.treesitter.require_language)
    lang))
 
+(fn valid-str? [lang code]
+  (-?> (vim.treesitter.get_string_parser code lang)
+       (#(doto $1 (: :parse)))
+       (: :trees)
+       (a.get-in [1])
+       (: :root)
+       (#(not ($1:has_error)))))
+
 {: enabled?
  : parse!
  : node->str
@@ -215,4 +223,5 @@
  : node-surrounded-by-form-pair-chars?
  : node-prefixed-by-chars?
  : get-form
- : add-language}
+ : add-language
+ : valid-str?}
--- a/fnl/conjure/client/guile/socket.fnl
+++ b/fnl/conjure/client/guile/socket.fnl
@@ -110,7 +110,8 @@
 (fn M.eval-str [opts]
   (with-repl-or-warn
     (fn [repl]
-      (let [context (or opts.context default-context)]
+      (if (ts.valid-str? :scheme opts.code)
+       (let [context (or opts.context default-context)]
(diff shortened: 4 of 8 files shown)


russtoku:
> Incomplete expressions will leave the Guile and scheme REPLs waiting for continued input effectively breaking it until it is disconnected/reconnected.

I notice that using `<localleader>ei` to interrupt the REPL doesn't work with the Guile REPL but does with the MIT Scheme REPL. Trying to interrupt the Guile REPL hangs Neovim in the code buffer.

However, disconnecting with `<localleader>cd` and connecting with `<localleader cc` does keep things working in the code buffer and it doesn't loose any previous work done in the Guile REPL.

seamusriordan:
Ah, this interrupt command is new to me.  I don't see an interrupt function defined for guile but I do see scheme's is sending a sigint directly to the process.  Since Guile is potentially a remote network connection we don't have such a luxury directly to the process.  I did a quick check to see if disconnecting would interrupt a long eval for

```
(let ((x 0)) 
  (while #t
         (set! x (+ 1 x))))
```

and it kept eating CPU cycles after the disconnect, so it's enough to get back to a prompt but unfortunately not a true interrupt.

As an aside, Guile does support multiple connections to the same environment as well as spawning servers in the REPL itself and I've been wondering if developing a different connection architecture might be the right approach.  For example something like having a "primary" control connection for monitoring/controlling/interrupting "secondary" connections where the eval is done.

seamusriordan:
> > As an aside, Guile does support multiple connections to the same environment as well as spawning servers in the REPL itself and I've been wondering if developing a different connection architecture might be the right approach. For example something like having a "primary" control connection for monitoring/controlling/interrupting "secondary" connections where the eval is done.
> 
> I found this hint: https://forum.systemcrafters.net/t/question-about-the-repl-workflow/850/2
> 
> which points to: https://www.gnu.org/software/guile/manual/html_node/Cooperative-REPL-Servers.html

Yes, this seems very much in that vein.  Something to investigate and tinker with

fnl/conjure-spec/client/guile/socket_spec.fnl
```diff
@@ -3,8 +3,12 @@
 (local guile (require :conjure.client.guile.socket))
 (local config (require :conjure.config))
 (local fake-socket (require :conjure-spec.client.guile.fake-socket))
+(local ts (require :conjure.tree-sitter))
 (require :conjure-spec.assertions)
 
+(tset package.loaded "conjure.remote.socket" fake-socket)
+(tset ts :valid-str? (fn [_ _] true))
```
Olical, line 10: A novel approach to mocking things, my worry is that we don't put these back to how they were at the start. So it can cause inconsistent test errors because they might run in different orders and step on each others toes.

So we could either put things back the way they were at the end of this test or (my personal choice) somehow pass in overrides to the functions in question, an optional dependency injection of a sort. Thoughts?

fnl/conjure/tree-sitter.fnl
```diff
@@ -200,6 +200,14 @@
        vim.treesitter.require_language)
    lang))
 
+(fn valid-str? [lang code]
+  (-?> (vim.treesitter.get_string_parser code lang)
+       (#(doto $1 (: :parse)))
+       (: :trees)
+       (a.get-in [1])
+       (: :root)
+       (#(not ($1:has_error)))))
```
Olical, line 209: I'd rather have two steps here, a function that parses and returns the root, another that checks if the tree is valid. This way we get two clear functions that can be reused and tested independently. Might also be worth breaking some of this out into various `let` assignments that we can name instead of a threading macro with these sort of anonymous function dances.

I see why they're required I think, but would be nice to make this syntax a liiiitle simpler. Sometimes when Lua APIs make it awkward to write in Fennel I break things out more to avoid using clever tricks to use the `:` syntax etc. It might be more verbose, but I'd rather things erred on the verbose side.

Olical (changes requested):
A good change I would just like to break that parser function up and label it's parts (it's okay if it takes more lines if it is clearer to read). And I would love a nicer way to override functions rather than mutating the module tables in a sort of mocking / stubbing style.

I would much rather the function takes extra args to override dependencies through a call instead of overriding things by mutating global state. If there's NO way to do this though, let's at least put the state back to how it was in the first place when the tests are done.

I'm pretty sure busted supports first class fixtures too?

fnl/conjure-spec/client/guile/socket_spec.fnl
seamusriordan: This is very reasonable and not resetting is a good concern.  Let me see what I can work up, especially if Busted has some functionality we can use.

fnl/conjure-spec/tree-sitter_spec.fnl
```diff
@@ -0,0 +1,60 @@
...
+
+(var saved-get-string-parser nil)
+
+(describe "conjure.tree-sitter"
+  (fn []
+    (describe "valid-str?" 
+      (fn []
+        (before_each
+          (fn []
+            (set saved-get-string-parser vim.treesitter.get_string_parser)))
```
seamusriordan, line 12: `vim.treesitter.get_string_parser` isn't expected to work in the testing environment (it doesn't seem to load any of the grammars) but I'll save/restore it around these specific tests.

fnl/conjure/client/scheme/stdio.fnl
```diff
@@ -114,53 +120,41 @@
...
- : stop
- : start
- : interrupt
- : on-load
- : on-filetype
- : on-exit }
+(fn M.on-exit []
+  (M.stop))
+
+M
```
Olical, line 160: Excellent work, thanks for doing this. Future changes will now be easier and more consistent with better interactive evaluations 🎉

Olical (approved):
Looks good now! Nice work!

Olical:
Tested locally, works well, prevented me from evaluating invalid code! Time to merge!

PR #688, opened 2025-07-15
Prevent Guile REPL completions when eval is pending

seamusriordan:
If the Guile REPL has a pending operation the REPL-based completions cannot not process resulting in nvim hanging. 

This PR places a check around the REPL state's `:current` state which represents the running evaluation.

Evaluating the following as a form or buffer will place the REPL in a state where it cannot process completions for 5 seconds, after which it will resume.

```
(sleep 5)
```

This branch will continue to operation (sans completions) until the evaluation completes or the REPL is disconnected/reconnected.


Olical:
Thank you very much!

PR #694, opened 2025-08-01
Fix iw text object as iW to select school.lesson-6-message

gbroques:
Fix `iw` text object as `iW` to select `school.lesson-6-message` since `iw` doesn't select the dot `.` as it's not included in the definition of a vim "word".

image


Olical:
Good catch! This might actually vary depending on use of tree-sitter or not, I think `iW` is safer though and will work in both cases, regardless of how your text objects are defined. A "word" is quite a personal definition within Neovim with lots of ways to move the boundaries per language so erring on the side of caution is a good idea.

PR #695, opened 2025-08-08
Add tree-sitter based completions to Guile, Scheme, and Common Lisp

seamusriordan:
This is a bigger one but largely similar to what I've been showing before.  I'm happy to change, delete, or add parts of this and acknowledge it is ambitious.

Treesitter can be used to parse and determine completions based on a provided set of queries.  This PR adds

- REPL-less completions based on the source code and cursor position
- A basic tree sitter query system to tag scopes of variables
- Configuration options to disable those completions (with additions to the documentation)
- a resource system which loads and caches files for use (such as the treesitter query files).  Inspiration was taken from how tree sitter manages its own query files
- A base dictionary for the basic scheme clients as they don't have good LSP support to provide global environment symbols
- Unit testing (and normalizing on "mock" over "fake" nomenclature)
- Light refactoring and organizing of existing completion code
- Continuing to switch files to the module-based export approach

The treesitter queries are designed so that it should be extensible to other languages (or at least lisp-likes with lexical scoping).  These languages were targeted due to their missing or limited LSP support, so hopefully this adds some good missing functionality.

The queries are best effort based on the documentation and I did my best to implement everything I could find within reason.  The common lisp queries in particular do not cover cases of things like autogenerated variables from classes and some of the more esoteric things like `define-method-combination`, but these are often provided via swank after evaluation.  Attached below are files that I built based mostly on examples from documentation that I used to help test various situations which you may find helpful in evaluating.

[scheme-completions-base.scm.txt](https://github.com/user-attachments/files/21692704/scheme-completions-base.scm.txt)
[lisp-completions-base.lisp.txt](https://github.com/user-attachments/files/21692705/lisp-completions-base.lisp.txt)

### DIFF
--- a/res/queries/common-lisp/cmpl.scm
+++ b/res/queries/common-lisp/cmpl.scm
@@ -184,7 +184,21 @@
     symbol: _ @_l)]
   . (list_lit
       (list_lit 
-        . (sym_lit) @local.define
+        . (sym_lit) @local.define)
+      @local.scope)
+  (#eq? @_pkg "cl")
+  (#any-of? @_l "flet" "labels" "macrolet"))
+@local.scope
+
+(list_lit
+  . 
+  [(sym_lit) @_l
+   (package_lit
+    package: _ @_pkg
+    symbol: _ @_l)]
+  . (list_lit
+      (list_lit 
+        . (sym_lit)
         . (list_lit (sym_lit) @local.bind)
         (#not-lua-match? @local.bind "^&.*")
         (#not-lua-match? @local.bind "^_$"))


russtoku:
Wow, waaay over my head! Quite awesome!! Can't wait to try it out.

russtoku:
Whew! Lots of cases to test.

```console
| Client      |  Enable completions? | REPL connected? | Result   |
|-------------|----------------------|-----------------|----------|
| Scheme      | true (default)       | Y               | works    |
| Scheme      | true (default)       | N               | works    |
| Scheme      | false                | Y               | no work  |
| Scheme      | false                | N               | no work  |

| Guile       | true (default)       | Y               | works    |
| Guile       | true (default)       | N               | works    |
| Guile       | false                | Y               | no works |
| Guile       | false                | N               | no works |

| Common Lisp | true (default)       | Y               | works    |
| Common Lisp | true (default)       | N               | no works |
| Common Lisp | false                | Y               | works    |
| Common Lisp | false                | N               | no works |
```
Is the Common Lisp case supposed to work like the Scheme and Guile cases?

seamusriordan:
The prior comments should be addressed and pushed up.  Those were all good suggestions.

russtoku:
The spec tests ran successfully.

russtoku:
Would it make sense to have the source of candidates (keyword, tree-sitter, and REPL-based) separated from each other? Then, have a completion function merge whatever is available.

I don't know enough about the completion candidate list to know if source of completion could be associated with the candidates so one could distinguish the source. There's also the question of which one to include if there are duplicates.

seamusriordan:
On second thought, keeping the completions toggle flag language-specific has some real use cases.  My motivation for doing this was Scheme, Guile, and Common Lisp not having great LSP support compared to other langs, so if an LSP were to become available I'd want to disable completions for only that language.

russtoku:
> Correct. I do observe the expected behavior locally so I might start with cross checking the configuration flags. I'm testing the "disabled" case with

Oh, I made a mistake! I didn't toggle the correct configurations. I need to redo my functional testing.

> Thinking about it, there's probably a case to be made to consolidate the enable completion flags into a single flag. 

Sorry, I wasn't suggesting this. I was thinking about functions internal to the clients' code; just to ease low-level debugging. From the user's perspective, I wouldn't want to have too many knobs to twiddle.

seamusriordan:
No, this is great and I appreciate the attention.  It is a lot of cases to go over.

I'm curious about this Common Lisp case where you're not getting any completions without the REPL.  Is it consistently on any file with lisp code in it?  Tree sitter will only pick up what it can parse and what it believes are variables it found in scope.  You might not get suggestions if it is at the very top of the file for example.

russtoku:
Yes, you are correct about the Common Lisp completions.

Using the `dev/common-lisp/sandbox.lisp` file, with completions for Common Lisp enabled and without the REPL connected, I get completion candidates that are function names but only if they are in the source lines above where the cursor is.

```console
  | Client      | Compl? | REPL? | Result | Expected | Pass |
  |-------------|--------|-------|--------|----------|------|
  | Common Lisp | true   | Y     | Y m    | Y        |  ✅  |
  | Common Lisp | true   | N     | Y s    | Y        |  ✅  |
  | Common Lisp | false  | Y     | N      | N        |  ✅  |
  | Common Lisp | false  | N     | N      | N        |  ✅  |
```

Olical:
I will also add that I love the amount of added tests, nice one!

fnl/conjure/util.fnl
```diff
@@ -1,12 +1,39 @@
...
+        result []]
+    (each [_ v (ipairs l)]
+      (when (not (. seen v) )
+        (tset seen v true)
+        (table.insert result v)))
+    (each [_ v (ipairs r)]
+      (when (not (. seen v) )
+        (tset seen v true)
+        (table.insert result v)))
+    result))
```
Olical, line 27: You might be able to replace this with something like `->set` https://github.com/Olical/nfnl/blob/66409b6186a9e313f60f4690b0f46fa2da1ee476/fnl/nfnl/core.fnl#L457-L465 where a set is a table under the hood. But up to you, I just get a bad vibe off of this function being two things at once. I would personally prefer two functions called `concat` and `distinct` (which are the function names Clojure uses) even if it's not quite as efficient execution-wise.

This function could be a chain of, `concat`, `->set`, `vals` from `nfnl.core`. Oh and `distinct` already exists in `nfnl.core`! So you could use `concat` + `distinct` from `nfnl.core` 😄

fnl/conjure/client/common-lisp/swank.fnl
```diff
@@ -8,10 +8,14 @@
 (local str (autoload :conjure.aniseed.string))
 (local text (autoload :conjure.text))
 (local ts (autoload :conjure.tree-sitter))
+(local cmpl (autoload :conjure.client.common-lisp.completions))
+(local util (autoload :conjure.util))
 
-(local buf-suffix ".lisp")
-(local comment-prefix "; ")
-(local form-node? ts.node-surrounded-by-form-pair-chars?)
+(local M (define :conjure.client.common-lisp.swank))
```
Olical, line 14: Thanks for performing this migration while you're here!

fnl/conjure/client/scheme/completions.fnl
```diff
@@ -0,0 +1,25 @@
...
+    (string.match command "mit") :mit
+    (string.match command "petite") :chez
+    (string.match command "csi") :chicken
+    :common))
+
+(fn M.get-completions [prefix]
+  (let [stdio-command (config.get-in [:client :scheme :stdio :command])
+        lang-key (get-lang-key-from-stdio-command stdio-command)
+        keyword-set (keywords.get-set lang-key) 
+        prefix-filter (util.make-prefix-filter prefix)]
```
Olical, line 20: I feel like this prefix-filter needs either renaming or breaking out in such a way that it's behaviour is more obvious, but I'm unsure. Maybe this is fine, it just feels like a REALLY specific function to this task, like maybe it needs to go into `conjure.tree-sitter-completions` which would make it feel a bit more at home? `conjure.utils` when it'll only really be used for this task (I think?) feels a bit too general.

fnl/conjure/resources.fnl
```diff
@@ -0,0 +1,29 @@
...
+
+(local resource-prefix "res/")
+
+(var cache {})
+
+(fn read-and-cache-file-contents [path]
+  (log.dbg (.. path " resource not cached - reading"))
+  (let [content (a.slurp path)]
+    (tset cache path content)
+    content))
```
Olical, line 15: Oh is this reading the TS query code into memory?

fnl/conjure/tree-sitter-completions.fnl
```diff
@@ -0,0 +1,120 @@
+(local {: autoload : define} (require :conjure.nfnl.module))
+(local a (autoload :conjure.nfnl.core))
+(local ts (autoload :conjure.tree-sitter))
+(local util (autoload :conjure.util))
+(local res (autoload :conjure.resources))
+
+(local M (define :conjure.tree-sitter-completions))
+
+(local symbol-query-path-template "queries/%s/cmpl.scm")
```
Olical, line 9: Interesting! Would probably be great to move the Clojure preamble code out of the client and into `.clj` files and lean on this. Really neat.

fnl/conjure/tree-sitter-completions.fnl
```diff
@@ -0,0 +1,120 @@
+(local {: autoload : define} (require :conjure.nfnl.module))
```
Olical, line 1: Very interesting module overall, not as scary as I thought it'd be thanks to the Scheme query syntax. Pretty cool! I hope it'll be okay to maintain into the future, I have a lot to learn around the query syntax to do so when issues eventually arise, but it's a very interesting idea.

Olical (changes requested):
Overall, good changes all over the place, nice migrations to the new way of doing modules. The query system is powerful and expressive so we don't have loads of manual tree walking and parsing like I feared.

I do still have this feeling that if this were a standalone project you could have a much broader scope of users because anyone who wants this also needs to buy into Conjure, which is maybe fine, but just an observation. The feature set could stand on it's own two legs I think, but maybe it'd be a lot of code to support the languages and the scaffolding around config etc, so maybe Conjure is the right home. There's no perfectly right answer there.

Thank you for putting all of this effort in and cleaning up while you go along, all the added tests are a wonderful addition too. I had a few comments around naming, function locations and that we might be re-inventing some `nfnl.core` ones, but nothing major.

I think I just need to give this a good test locally to build confidence and make sure there's no weird edge cases or errors.

seamusriordan:
> Going to take a look through this today, my main concern is that it's a lot more new code and it's kinda stepping outside the scope of Conjure.
> ...
> But this division of workloads is in an ideal world, Conjure already blurs various lines in various languages, so what's a little more blur on top at the end of the day? I still want to communicate my thoughts on roles and what should be doing what in the tooling ecosystem for future reference 🙂

This is a valid concern, one I shared as I was working through it and watching it grow, and I appreciate your directness.  If we decide conjure isn't the right place I expect I'll extend it to a separate plugin without too much trouble, it does sit in an awkward place between the editor and an LSP.

fnl/conjure/util.fnl
seamusriordan: Excellent and I think that's a good idea.  Just my own unfamiliarity with nfnl.

seamusriordan:
> Maybe one day you can spin this out into your own standalone plugin and we can slowly point people at that since it'll be highly specialised and probably more powerful by then, but I do think that Conjure's platform is a good place to start, so I'm not completely averse. I just get more cautious the larger the PR gets 😅

Good enough, just let me know what you decide.  Going through the prior comments if smaller parts like the resource system and the module refactoring done are things we find useful regardless I can cherry pick those out.

Olical:
I'm still happy for this PR to go forward btw, I'm just thinking for the future, if you want to develop this idea in a standalone system then I'd totally support that too. Sorry if that's unclear, I'm just thinking carefully out loud about the changes is all, I do like the whole thing overall, I was just nervous about taking on more scope for the project since it's _quiiiite_ broad anyway. Conjure is a bit of a kitchen sink these days, but maybe that's fine!

seamusriordan:
Very good and we're on the same page (I've done my share of thinking out loud on this too).  I'll propose something for the `prefix-filter` comment and rework the deduplication with `distinct` hopefully shortly.

fnl/conjure/util.fnl
```diff
@@ -1,12 +1,25 @@
...
   (fn []
     ((. (require mod) f))))
 
-(fn replace-termcodes [s]
+(fn M.replace-termcodes [s]
   (vim.api.nvim_replace_termcodes s true false true))
 
-{: wrap-require-fn-call
- : replace-termcodes}
+(fn M.ordered-distinct [l]
```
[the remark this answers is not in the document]
Olical: It's fine here for now, I like the implementation, but we could totally port this over into the nfnl repo at some point soon, will be a solid improvement I think.

Olical (approved):
I'm happy with this now! As we discussed, it's a big addition and it may find another home dedicated to this functionality some day, but this is a good home for now. Great work! I'll just have a play with it locally before merging.

PR #698, opened 2025-08-19
TypeScript support for JavaScript client

Stansom:
Hello! I’ve added TypeScript support for the JavaScript client, and it works pretty well, but I have a question: must it be presented as a separate client, or is it okay to be the same client as JavaScript?

There are a few quirks, such as:
1. ts-node yells at an unexpected “export” token, but a simple re-evaluation helps.
2. Not all import statements are currently supported. It requires something more complex in terms of parsing and transcribing imports into require statements.
3. The file evaluation logic was changed. Now, it takes an opened file, makes some changes, saves it as a temporary file, and passes it into the REPL using the .load flag to be evaluated as a single buffer. After that, the temporary file is removed. However, with this method, there is no direct output into the REPL, and it goes into a "stray" output. Because of this behavior, it is better to have "stray" output activated for JavaScript/TypeScript when evaluating an entire file.


russtoku:
> must it be presented as a separate client, or is it okay to be the same client as JavaScript?

I think that it's OK to have one client for both JavaScript and TypeScript.

Nicely done!

Olical:
Rebasing on main should fix the CI too, there were some issues with the version tags I was using which I've since fixed by specifying explicit Neovim versions for the tests to run on.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -10,10 +10,26 @@
 
 (local M (define :conjure.client.javascript.stdio))
 
+(fn filetype [] vim.bo.filetype)
+
+(local repl-type (if (= "javascript" (filetype))
+                     :js 
+                     
+                     (= "typescript" (filetype))
+                     :ts))
```
Olical, line 19: This locks the REPL type in when this modules loads. We probably want it more dynamic than that so that the user can restart the REPL in a different file and get the appropriate REPL started.

I'd consider converting `repl-type` to a function that we call multiple times just to make sure it's relevant to the current `vim.bo.filetype` instead of the one we had open when the module loaded.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -51,60 +69,90 @@
 (fn display-result [msg]
   (log.append msg))
 
-(fn replace-require-path [s cwd]
-  (if (string.find s :require)
-      (string.gsub s "require%(\"(.-)\"%)"
+(fn get-absolute-path [f]
+  (.. "\"" 
+      (vim.fn.fnamemodify (.. (vim.fn.expand "%:p:h") "/" f) ":p")
```
Olical, line 74: I have a feeling this `/` will have a bad time on Windows, we might need to use `vim.fs.normalize` or `conjure.fs/path-sep` maybe which is OS dependant.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -51,60 +69,90 @@
...
+        replace-fn (fn [bd path]
+                     (let [spl (str.split bd ",")
+                           spl->nms (->> 
+                                      spl 
+                                      (a.map (fn [el] (el:gsub "as" ":")))
+                                      (str.join ", "))]
+
+                       (.. "const {" spl->nms "} = require(\"" path "\")")))
+        (repl _) (string.gsub s pattern replace-fn)]
+    repl))
```
Olical, line 100: Unsure exactly how these two are working but I'll take your word for it / trust how it behaves in the sandbox. Seems like hard stuff to get right! I'm guessing there will always be edge cases but we can work on those over time.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -51,60 +69,90 @@
...
+        final-acc (a.reduce 
+                    (fn [acc [pat repl]]
+                      (if acc.applied?
+                          acc
+                          (let [(r c) (string.gsub acc.result pat repl)]
+                            (if (> c 0)
+                                {:applied? true :result r}
+                                acc))))
+                    initial-acc patterns-replacements)]
+    final-acc.result))
```
Olical, line 125: Something we may be able to do which might make a lot of these find and replaces go away is use acorn in the REPL as shown in the issue linked from one of the comments here https://github.com/nodejs/node/issues/48084#issuecomment-1557952691

Basically we'd first send the JS code off as a string into the REPL, use the acorn functions to unparse it into something the REPL could work with then send it back in to be evaluated.

Not something we need to do right now, but if edge cases keep coming up, maybe using node to rewrite the code into something the REPL is okay with would work better than trying to translate it at the source code string level ourselves.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -113,34 +161,55 @@
...
+        sub-fn (fn [ln]
+                 (if (or (text.starts-with ln :.)
+                         (string.match ln "%s*@")
+                         (text.ends-with ln "{")
+                         (text.ends-with ln ";")
+                         (str.blank? ln))
+                     ln
+                     (.. ln ";")))
+        sub (a.map sub-fn spl)]
+    (str.join " " sub)))
```
Olical, line 186: As I said in the other comment, might be a good idea to always append `;` if it doesn't hurt. Or send another eval after each one with `;` in it and nothing else just to reset things.

fnl/conjure/client/javascript/stdio.fnl
```diff
@@ -169,10 +238,28 @@
        (a.map prepare-out)
        (str.join "")))
 
+(fn delete-file [f]
+  (let [cmd (if (= 0 (vim.fn.has "macunix"))
+                "del"
+                "rm")]
+    (when (= 1 (vim.fn.filereadable f))
+      (os.execute (.. cmd " " f)))))
```
Olical, line 246: Fairly certain we can use `vim.uv.fs_unlink` instead which is cross platform.

Olical (changes requested):
Some changes around OS compat stuff, but it looks pretty good otherwise!

fnl/conjure/client/javascript/stdio.fnl
Stansom: I redesigned the import replacement methods and added tests to make it easier to catch/fix edge cases.

fnl/conjure/client/javascript/stdio.fnl
Olical: And I'm okay with either, I couldn't actually find `rm` and `fs_unlink` just stood out to me. I don't mind which one is used, mostly just wanted an API that paves over OS differences.

fnl/conjure/client/javascript/stdio.fnl
Olical: Ah okay, that's fine, my thinking though is that `;` is whitespace and not essential but it helps fix certain TypeScript cases then we should default to appending it because:

 - It fixes all cases where `;` is required
 - It doesn't cause issues in cases where it's optional
 - We can delete code that checks for certain cases, less code, more robust

If any of that doesn't hold true though, please feel free to ignore! I just like to find ways we get to be "lazy" but the outcome is: Less code to maintain, more consistent behaviour with less chance for edge cases.

Maybe appending `;` when it's not required could cause issues, but from what I understand it's whitespace and used to separate expressions. I used to get away with no semi-colons but I know doubling them up is also fine.

fnl/conjure/client/javascript/stdio.fnl
Stansom: You commented on an old commit, now the repl-type is a function.

fnl/conjure/client/javascript/stdio.fnl
Stansom: Unfortunately, we have to be selective here because `;` may not be suitable everywhere. For example, they will ruin everything when defining interfaces. I hope that the future Node.js REPL will support TypeScript better, and we won't have to deal with all these manipulations at all.

fnl/conjure/client/javascript/stdio.fnl
Stansom: I think I'll stick with `vim.uv.fs_unlink` because it's supported by older versions of Neovim.

Olical:
I'll take another look asap but pushed some changes to https://github.com/Olical/conjure/tree/Stansom-main for now, getting off my train now so can't double check and finish. I just simplified things and remove the statefulness of the command.

If we want this to be configurable I'd create a typescript-command and javascript-command config option which we select based on filetype at invocation time rather than having to call another function to update the state at certain times.

I think I'm happy with this branch + my few commits on top, please feel free to take a look and let me know your thoughts too. I also made use of the `args` config since that was sitting unused and `--experimental-repl-await` was never being passed to `node`. I've fixed that.

Stansom:
> I'll take another look asap but pushed some changes to https://github.com/Olical/conjure/tree/Stansom-main for now, getting off my train now so can't double check and finish. I just simplified things and remove the statefulness of the command.
>
> If we want this to be configurable I'd create a typescript-command and javascript-command config option which we select based on filetype at invocation time rather than having to call another function to update the state at certain times.
> 
Oh yes, that's what I didn't like either, but I couldn't understand why. 😊 Thank you for showing me.
> I think I'm happy with this branch + my few commits on top, please feel free to take a look and let me know your thoughts too. I also made use of the `args` config since that was sitting unused and `--experimental-repl-await` was never being passed to `node`. I've fixed that.
>
I thought that the handler somehow automatically adds arguments from the table passed in the configuration to the command field.  😅

Everything looks pretty good, I think. Perhaps the only thing to do is make the -i flag non-modifiable by a user, because removing it could break both REPLs.

Stansom:
@Olical Oh, sorry. I probably made a mistake when I merged from the upstream repository. I fixed it, though, and everything looks good now.

Olical:
So I can get this working with `node` and `.js` files but the `.ts` files don't yield any output, is that the same for you?

Also I pushed a commit to your branch that makes it use `npx` so I don't need `ts-node` installed globally, but feel free to remove that if you disagree!

Stansom:
> So I can get this working with `node` and `.js` files but the `.ts` files don't yield any output, is that the same for you?
> 
> Also I pushed a commit to your branch that makes it use `npx` so I don't need `ts-node` installed globally, but feel free to remove that if you disagree!

Very interesting... Everything works fine for me with my local setup and Neovim 0.11.4. However, it's not working in a Docker container with Neovim 0.12. The problem is that neither <localleader>ee nor <localleader>er are working, but the <localleader>E motion evaluation is working. Are there some tree-sitter problems? I don't really know; it needs some investigation.
@russtoku , could you please check on your machine?

russtoku:
Everything in `dev/typescript/sandbox.ts` evaluates as expected with `<localleader>ee`, `<localleader>er`, `<localleader>ew`, and `<localleader>E` (normal and visual modes).

Very nice!

I'm running:
- Conjure from commit bedb508c on the main branch of https://github.com/Stansom/conjure/
- Neovim 0.11.3
- Using built-in package handling; no package manager like lazy, Packer, etc.
- nvim-treesitter - main branch
- Node.js 22.19.0 via mise
- npx - 11.6.0 via mise
- ts-node 10.9.2
- macos 15.6.1 Sequoia on M3 Macbook Air 15"

russtoku:
Very nice! Your code reads a lot clearer.

Olical:
So I'm using the same version of ts-node but when I use it on the CLI I get output like this:


```
ts-node
> 10 + 20;
| 
```

It just prints pipes and doesn't actually evaluate. I have a feeling that's why I'm also getting no output in Conjure.


```
// Sponsored by @Ajetski ❤
// WARNING! Node.js REPL limitations require transformations:
// 1. ES6 'import' statements are converted to 'require(...)' calls.
// 2. Arrow functions ('const fn = () => ...') are converted to 'function fn() ...' declarations to allow re-definition.
// --------------------------------------------------------------------------------
// "ts-node -i" (started)
// debug: remote.stdio.next-in-queue; stdin:write next-msg.code >>"\n"<<
nil
// debug: remote.stdio.on-message; receive source >>out<<
nil
// debug: remote.stdio.on-message; receive err >>nil<<
nil
// debug: remote.stdio.on-message; receive chunk >>"> "<<
nil
// debug: remote.stdio.on-message; receive source >>out<<
nil
// debug: remote.stdio.on-message; receive err >>nil<<
nil
// debug: remote.stdio.on-message; receive chunk >>"| "<<
nil
// --------------------------------------------------------------------------------
// eval (root-form): const add = (a: number, b: number) => a + b
// debug: remote.stdio.next-in-queue; stdin:write next-msg.code >>"function add(a: number, b: number) { return a + b }↵"<<
nil
// debug: remote.stdio.on-message; receive source >>out<<
nil
// debug: remote.stdio.on-message; receive err >>nil<<
nil
// debug: remote.stdio.on-message; receive chunk >>"| "<<
nil
// --------------------------------------------------------------------------------
// eval (root-form): add(38, 4)
// --------------------------------------------------------------------------------
// eval (root-form): 8 + 5;
// --------------------------------------------------------------------------------
// eval (root-form): (3 + 9 + 30) / 2;
// --------------------------------------------------------------------------------
// eval (selection): 8 + 5;
```

Trying to work out why, it seems like a me problem, but I don't know why just yet. I like the code changes though! Should be far more robust!

Olical:
Oh I just learned ts-node is no longer maintained and may have issues with the REPL support. I wonder what the recommended approach these days is, or maybe they don't have a REPL solution at all? And they just assume you'll compile on disk and run the JavaScript output with `node`? I saw someone mentioning that I might need to downgrade to `ts-node@9` but it's so old!

Olical:
Hm if I use `npx ts-node@9` it does work.

```
> 10 + 20;
30
```

So I have a horrible feeling ts-node is just completely broken for me and I don't see it ever being fixed. I wonder if there's an alternative TypeScript REPL project out there or if the TypeScript community has just moved away from REPLs entirely?

russtoku:
I have Node.js 22.19.0 installed on my M3 Macbook Air running Tahoe (macos 26.0). While the [documentation](https://nodejs.org/en/learn/typescript/run-natively) says it can handle TypeScript, I don't see that with this:

```TypeScript
function foo(bar: number): string {
  return 'hello';
}
```

```console
$ node -i
Welcome to Node.js v22.19.0.
Type ".help" for more information.
> function foo(bar: number): string {
function foo(bar: number): string {
                ^

Uncaught SyntaxError: Unexpected token ':'
>
```
It does work when I run `npx ts-node -i`. Although, it's necessary to evaluate some code like `1+1` before the function definition is evaluated without an error about "Unexpected token 'export'".

```console
$ npx ts-node -i
> function foo(bar: number): string {
... return 'hello';
... }
/Users/russ/Projects/Conjure/russ-try/PR-698/conjure/<repl>.ts:5
export {};
^^^^^^

Uncaught SyntaxError: Unexpected token 'export'
>
```

I'm running Node.js via `mise` and ts-node is at version 10.9.2.

```console
$ mise list node
Tool  Version  Source                                                Requested
node  22.19.0  ~/Projects/Conjure/russ-try/PR-698/conjure/mise.toml  latest

$ npx ts-node --version
v10.9.2
```

Stansom:
@russtoku 
> have Node.js 22.19.0 installed on my M3 Macbook Air running Tahoe (macos 26.0). While the [documentation](https://nodejs.org/en/learn/typescript/run-natively) says it can handle TypeScript, I don't see that with this:

Unfortunately, according to the documentation, the future is still experimental. For advanced usage, it is recommended to use ts-node: https://nodejs.org/en/learn/typescript/run

> Although, it's necessary to evaluate some code like 1+1 before the function definition is evaluated without an error about "Unexpected token 'export'"

Very good catch! I will try to evaluate a simple expression, such as 1 + 1, silently after starting the REPL, in order to avoid bothering the user with an error message.

russtoku:
Might I suggest the following changes:

```fennel
$ git diff fnl/conjure/client/javascript/stdio.fnl
diff --git a/fnl/conjure/client/javascript/stdio.fnl b/fnl/conjure/client/javascript/stdio.fnl
index 3dd24d44..ca777674 100644
--- a/fnl/conjure/client/javascript/stdio.fnl
+++ b/fnl/conjure/client/javascript/stdio.fnl
@@ -14,6 +14,8 @@
                {:javascript
                 {:stdio
                  {:args "-i"
+                  :javascript-command "node --experimental-repl-await"
+                  :typescript-command "ts-node"
                   :prompt-pattern "> "
                   :show_stray_out false}}}})

@@ -135,12 +137,13 @@
 (set M.initialise-repl-code "1+1")

 (fn repl-command-for-filetype []
+  (log.append [(.. M.comment-prefix "Filetype: "  vim.bo.filetype)])
   (if
     (= :javascript vim.bo.filetype)
-    "node --experimental-repl-await"
+    (cfg [:javascript-command])

     (= :typescript vim.bo.filetype)
-    "ts-node"))
+    (cfg [:typescript-command])))


 (fn M.start []
```

It adds:
- vim.g["conjure#client#javascript#stdio#javascript-command"]
- vim.g["conjure#client#javascript#stdio#typescript-command"]

to allow the user to use "npx ts-node" for the TypeScript command to run for the REPL. Of course, one could drop the `conjure#client#javascript#stdio#args` config item because it can be added to the command strings.

The log form provides some additional reassurance to the user that the REPL is for a TypeScript buffer.

Stansom:
@Olical 
I'm really glad to hear that everything's working for you now.
The note about how VimL handles naming conventions is super helpful and good to know for future projects.
I'll apply your changes locally and get everything ready for you.

daveyarwood:
That's a good theory - my default Node version is 24.

I tried `nvm use 22.16.0`, confirmed via `node --version`, and then opened Neovim in the same shell session, but I'm still seeing the same behavior. I tried `:!node --version` inside Neovim as a sanity check, it said 22.16.0.

daveyarwood:
Got it working! Reading the help file was essential. I didn't have `ts-node` installed globally, nor did I have a `tsconfig.json` in my test project.

I ran `npm install -g ts-node` and tried evaluating very simple snippets (e.g. `1 + 1`) and those work 🎉 

It's definitely not capable of running larger / more complex code excerpts at this stage, which is fine (gotta start somewhere!). I think the main obstacle for me is that in my company's code base, we never use semicolons, and it seems like this implementation relies on statements being semicolon terminated?

This is exciting, nonetheless. Awesome job to all of the contributors!

daveyarwood:
Some minor notes about the part of the help text that talks about `tsconfig.json`:

```
TypeScript support:
1. Install NodeJS(v16 - v22 tested, >22 do not work)
2. Install ts-node(v10.9.2 tested)
3. Create a minimal config file(tsconfig.json) in the root of a project:
   {
     "ts-node": {
       // ts-node config here
       }
     },
     "compilerOptions": {
         "target": "es2022",
         "module": "es2022"
     }
   }
4. Try it
```

The minimal tsconfig example does not work as is -- there's an extra curly brace, making it syntactically invalid JSON. It should be like this:

```javascript
{
  "ts-node": {
    // REQUIRED: Valid config must go here. Cannot be blank!
  },
  "compilerOptions": {
    "target": "es2022",
    "module": "es2022"
  }
}
```

I tried leaving the `"ts-node": { ... }` part empty and it didn't work. It seems that this is not a working minimal tsconfig.json, so we may want to document that better, or perhaps include a working minimal tsconfig.json right there in the help text that users can copy and paste into a test project.

Just some feedback, hope it's helpful!

daveyarwood:
Example errors I'm getting when I try to evaluate the first two lines in one of my company's TS files (for my own reference, it's `collectionSettings.ts`):

Evaluated code:

```typescript
import _ from 'lodash/fp.js'
```

Error:

```
// eval (selection): import _ from 'lodash/fp.j...
(out) Expression assignment to _ now disabled.
(out) undefined
```

Evaluated code:

```typescript
import {
  type ConvertOptions as CrateConvertOptions,
  type OptimizationOptions,
} from 'mongo2crate'
```

Error:

```
// eval (selection): import { type ConvertOptio...
(err) <repl>.ts(10,8): error TS2451: Cannot redeclare block-scoped variable 'type'.
(err) <repl>.ts(10,50): error TS2451: Cannot redeclare block-scoped variable 'type'.
(err) <repl>.ts(10,13): error TS1005: ',' expected.
(err) <repl>.ts(10,55): error TS1005: ',' expected.
```

If it's expected that snippets like these won't work at this stage, that's totally fine. Just wanted to report these results in case they're helpful.

Stansom:
> import _ from 'lodash/fp.js'

I just wanted to let you know that I checked on my machines, and it works. I'm not sure what the issue could possibly be, but it could be something with your project's settings or the configuration of your machine.

daveyarwood:
I realized today that having Conjure enabled was messing up my TypeScript LSP setup, so I ended up disabling Conjure for TypeScript files. I'm not sure if maybe that may have been related to the evaluation errors above, but at any rate, I'm fine to leave it there for now. Maybe I'll try again at some point in the future -- too much other work to do at the moment!

Thanks all, I appreciate the help.

Stansom:
> I realized today that having Conjure enabled was messing up my TypeScript LSP setup, so I ended up disabling Conjure for TypeScript files.

That's really strange. I don't think Conjure has anything to do with the TypeScript LSP, but I could be wrong. Thanks so much!

PR #705, opened 2025-09-02
fix error in issue #466

### LINKED ISSUE #466: Common Lisp Completion

bx-sa1:
Resolves #466


Olical:
Nope, just those in Conjure's source files. The .nfnl.fnl file is loaded by
the plugin at dev time so we will have nfnl installed.

Dev time stuff is fine, it's user space runtime things that need to refer
to the embedded one.

Issue #689, opened 2025-07-16
Try buffering log file output


Olical:
Rapid log appending from a tight loop of some kind can lock up Neovim, can buffering log output for a few ms help the performance?

## To Do

 - [x] Add benchmarks for bulk log appends
 - [x] Implement a generic buffering system, maybe in nfnl
 - [x] Add buffering and check if it helped

Olical:
Now I have a setup function that ensures the log is open and visible (even though it's headless?) before running the benchmark. I'm also doing 50 appends per iteration.

```
# bencode
## simple encode and decode x1000 [0.011ms]
## big encode, chunked decode x1000 [2.635ms]
# log
## one logging call at a time x1000 [0.041ms]
## 50 log calls in a row x100 [39.85ms]
```

Let's see what batching writes does to this.

Olical:
Implementation plan:

 - Migrate `log.append` to `log.direct-append`, that's what does the actual writing.
 - `log.append` will capture the client information and lines and add them to a buffer, maybe indexed by client somehow.
 - On startup a timer is created that will call `log.flush` on a configurable interval.
 - `log.flush` gathers up the data in the buffer and calls `log.direct-append` for each file type and then wipes the buffer.
 - We can only batch calls that use basically the same opts, if they vary so like some use `break?` or `suppress-hud?` then we might need to unbatch those calls. `supress-hud?` at least can be implemented with an `and`, if all calls use it then we toggle it on, if one of them does not use it then we might as well show the HUD because at least one message in there wants the HUD.
   - `break?` however does need to be smart, so maybe we can find a way to batch as much as possible in between the breaks. This is still useful because we're optimising for cases where a system may print out 1000s of lines of stdout with no break usage. Breaks are mostly only used by user input events, not streaming out from the program.

Olical:
I'll push this on a branch because it's a big change and I might have introduced bugs, but 40ms -> 1ms is quite the improvement 😄 

```
# bencode
## simple encode and decode x1000 [0.011ms]
## big encode, chunked decode x1000 [2.483ms]
# log
## one logging call at a time x1000 [0.047ms]
## 50 log calls in a row x100 [1.16ms]
```

PR #713, opened 2025-10-03
Implement batched logging with auto flush

Olical:
#689

## Before

```
# bencode
## simple encode and decode x1000 [0.011ms]
## big encode, chunked decode x1000 [2.635ms]
# log
## one logging call at a time x1000 [0.041ms]
## 50 log calls in a row x100 [39.85ms]
```

## After

```
# bencode
## simple encode and decode x1000 [0.011ms]
## big encode, chunked decode x1000 [2.483ms]
# log
## one logging call at a time x1000 [0.047ms]
## 50 log calls in a row x100 [1.16ms]
```

Win?


fnl/conjure/client.fnl
```diff
@@ -1,145 +1,133 @@
...
+(fn M.wrap [f ...]
+  (let [opts {filetype (core.constantly (filetype))
+              M.state-key (core.constantly (M.state-key))}
         args [...]]
     (fn [...]
-      (if (not= 0 (a.count args))
-        (dyn.bind opts f (unpack args) ...)
-        (dyn.bind opts f ...)))))
+      (if (not= 0 (core.count args))
+          (dyn.bind opts f (unpack (core.concat args [...])))
```
Olical, line 71: The linter was complaining saying only the first value from `args` would be used in the original code. So this change _should_ fix that subtle bug and mean that it uses all of the args now. Unsure though! I'll write a test.

PR #719, opened 2025-10-31
fix(common-lisp): make doc_word working for symbol which has package

lost22git:
prefix, like: `(json:parse)`


Olical:
Makes sense to me, I'll take your word for it that it's required 😄 thanks!

PR #721, opened 2025-11-05
Complete migration from Aniseed to nfnl.

russtoku:
I think this PR is the last step to complete the migration from Aniseed to nfnl. Hopefully, it's not too premature.

It:
- Migrates `conjure.remote.transport.netrepl` to nfnl reloadable module.
- Removes the Aniseed-based Fennel client code and its help doc.
- Removes mention of Aniseed from the Conjure help doc.
- Removes tests for Aniseed and dependency of the test script on Aniseed.
- Removes the vendored Aniseed modules.

`scripts/test` passes all tests. I successfully evalutated the sandbox code for Python and Scheme (using Chez Scheme). I don't use Janet so I'm trusting that the tests for the netrepl transport assures that the Janet client still works.

### DIFF
--- a/scripts/setup-test-deps
+++ b/scripts/setup-test-deps
@@ -16,12 +16,6 @@ else
   git clone https://github.com/bakpakin/fennel.vim.git "$PACK_DIR/fennel.vim"
 fi
 
-if [ -d "$PACK_DIR/aniseed" ]; then
-  echo "aniseed already exists"
-else
-  git clone https://github.com/Olical/aniseed.git "$PACK_DIR/aniseed"
-fi
-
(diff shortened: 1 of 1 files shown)


Olical:
So I'm all for the conversions and removing some doc references but I'm inclined to keep the Aniseed module and it's tests around for the foreseeable future, probably forever. Some users may have pinned their Fennel client to the Aniseed one and they may have no intention of migrating away from it.

I don't mind having the client and the library buried in the project, we just need to not require those modules anymore unless a user explicitly still wants to use Aniseed for whatever reason.

I'll pull your branch down locally and put a few files back, but otherwise the change is good!

PR #729, opened 2025-11-17
Fix(javascript): Handle namespace imports correctly

Stansom:
This change addresses a bug in how JavaScript namespace imports (e.g., import * as name from "module") are transformed.

The previous implementation incorrectly assumed the structure of the AST node for a namespace import, leading to errors when trying to get the alias. This has been corrected by using tsc.find-child-by-type to reliably find the identifier node.

This ensures that namespace imports are correctly converted to require statements.

### DIFF
--- a/fnl/conjure/client/javascript/transformers.fnl
+++ b/fnl/conjure/client/javascript/transformers.fnl
@@ -3,6 +3,7 @@
 (local tsc (autoload :conjure.client.javascript.ts-common))
 (local ir (autoload :conjure.client.javascript.import-replacer))
 (local log (autoload :conjure.log))
+(local a (autoload :conjure.nfnl.core))
 
 (local M (define :conjure.client.javascript.transformers))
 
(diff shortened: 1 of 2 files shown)


Olical:
Tried out the before and after locally, works fine for me and the error is printed nicely.

fnl/conjure/client/javascript/transformers.fnl
```diff
@@ -34,18 +46,42 @@
...
+              (= ntype "super")
+              (= ntype "meta_property") 
+              (and (= ntype "identifier") (= (tsc.get-text n code) "arguments"))
+              (= (tsc.get-text n code) "new.target")) 
+          (set found n)))
+      (when (not found)
+        (each [child (n:iter_children)]
+          (traverse child)))))
+  (traverse node)
+  found)
```
Olical: Instead of using a `var` and mutating it I'm pretty sure this can be implemented with an `if` statement and returning the result of `(traverse node)` - inside the recursive function we can return at the point you `(set found n)` . We don't actually use a `return` keyword though which breaks out of the control flow, instead we should have two steps.

In the `let` block we can do the lookup, if we find a value, we end the recursion and return the found value in the tail position. I think we also have to check the result of each sub-call to `traverse` in that `each` so we return if something is returned from the child call.

I think this current code actually iterates more than it needs to and doesn't exit early when it finds a match? Just a small optimisation but I think this can be written in a true recursive (although maybe not tail recursive because of the tree nature, that would require more state to be passed along in the recursion which I don't care about so much) function that would be easier to follow and perform less work when it its an error.

I could be wrong about this though because I've been pretty sick all week and I'm still kind of out of it today but I wanted to get a response to you 😅 

If you disagree or would rather not try to implement this in a recursive way, please let me know, I'm open to being swayed.

Issue #736, opened 2025-12-12
Olical blog changes means post link in README no longer valid


kittywitch:
Noticed in README, links to https://oli.me.uk/conversational-software-development/
No longer exists, instead https://oli.me.uk/Blog+archive/2020/Conversational+software+development is correct

PR #737, opened 2025-12-12
Fix link to post in README

kittywitch:
Resolves https://github.com/Olical/conjure/issues/736


Olical:
Oh!? I went to merge yours thinking it was another broken link and just realised I just merged another one for the same link. What is that timing! So, uh, this is applied but from another coincidental PR from within < 12 hours of each other, that's odd. Thank you for this though, it's a good catch and the old link should really work, I will update my Obsidian thing so the old link continues to work, even though this fix is also correct.

Thanks!

Olical:
Ah this DID work, but Obsidian updated and now it's `aliases` not `alias`. I just need to update the name of the property where ever I used it.

PR #744, opened 2025-12-31
Add Ruby client

russtoku:
This PR adds a Conjure client for the Ruby programmaing language.

It's been developed with Ruby version 3.4 and should also work with the  just released version 4.0.
You'll probably be able to use it with Ruby on Rails.


Olical:
Worked well for me! Nice one!

PR #750, opened 2026-01-14
Avoid pcall on vim.treesitter.language.add as it doesn't assert/error

jship:
`pcall`-ing `vim.treesitter.language.add` would always return true as its first result value, as `add` uses the style of returning nil and an error message when it can't add the language rather than aborting via `error` or `assert`. This causes an issue when a Conjure user opens a Python buffer but doesn't have a Python treesitter parser installed. Conjure thinks they _do_ have a parser installed and tries to start the stdio client. The user would see an error from Neovim's treesitter code like `vim.schedule: No parser for language "python"`, and not the stdio client's intended log messages for this case (reproduced below) which made it difficult to identify that the error was originating from Conjure.

```
(error) The python client requires a python treesitter parser in order to function.
(error) See https://github.com/nvim-treesitter/nvim-treesitter
(error) for installation instructions.
```

The style of return nil plus error message is a change from the now-deprecated `require_language` functions that Conjure's `add-language` also wraps for backwards-compatiblity. The `pcall` was previously around the `add-language` wrapper call in the Python stdio client's `start` function. These changes move the choice to `pcall` or not into the `add-language` wrapper itself. This lets `add` skip a `pcall` while the deprecated `require_language` functions still get a `pcall`.

### DIFF
--- a/fnl/conjure/tree-sitter.fnl
+++ b/fnl/conjure/tree-sitter.fnl
@@ -190,10 +190,11 @@
             node))))))
 
 (fn add-language [lang]
-  ((or vim.treesitter.language.add
-       vim.treesitter.language.require_language
-       vim.treesitter.require_language)
-   lang))
+  (let [add (case vim.treesitter
+              {:language {:add f}} f
+              {:language {:require_language f}} (partial pcall f)
+              {:require_language f} (partial pcall f))]
+    (add lang)))
 
 (fn get-root-node-for-str [lang code]
   (let [parser (vim.treesitter.get_string_parser code lang)]
--- a/fnl/conjure/client/python/stdio.fnl
+++ b/fnl/conjure/client/python/stdio.fnl
@@ -237,7 +237,7 @@
                      (config.get-in [:mapping :prefix])
                      (cfg [:mapping :stop]))]
                 {:break? true})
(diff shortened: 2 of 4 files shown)


fnl/conjure/tree-sitter.fnl
```diff
@@ -190,10 +190,11 @@
...
 
 (fn add-language [lang]
-  ((or vim.treesitter.language.add
-       vim.treesitter.language.require_language
-       vim.treesitter.require_language)
-   lang))
+  (let [add (case vim.treesitter
+              {:language {:add f}} f
+              {:language {:require_language f}} (partial pcall f)
+              {:require_language f} (partial pcall f))]
```
russtoku: It doesn't appear that the clause on L.196 would ever match because `:require_language` isn't a key in `vim.treesitter`.

Olical: Nice, I would also point out that our minimum version is `v0.10` so we should try to be backwards compatible to there but if someone's on `<0.10` Conjure will error anyway. I didn't look into the version numbers these changes correspond to but I'm okay with this backwards compatibility regardless, just thought I'd mention the version threshold anyway.

I also didn't know Fennel case statements could be used like this, I really like it!

PR #751, opened 2026-01-14
Python client was doing some on-success init inside call to display-repl-status

jship:
In the Python stdio client's `on-success`, the `with-repl-or-warn` call was accidentally within the `display-repl-status` call (`display-repl-status` is only meant to take a single status arg). Looking at the previous generated Lua, this wasn't causing a problem as the `with-repl-or-warn` block was still getting called. These changes are more to avoid future confusion and mixups.

### DIFF
--- a/fnl/conjure/client/python/stdio.fnl
+++ b/fnl/conjure/client/python/stdio.fnl
@@ -250,8 +250,8 @@
 
            :on-success
            (fn []
-             (display-repl-status :started
-              (with-repl-or-warn
+             (display-repl-status :started)
+             (with-repl-or-warn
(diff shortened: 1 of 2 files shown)


fnl/conjure/client/python/stdio.fnl
```diff
@@ -260,7 +260,7 @@
                  (repl.send
                    (prep-code M.initialise-repl-code)
                    (fn [msgs] nil)
-                   nil)))))
+                   nil))))
```
Olical: Good catch!

Olical (approved):

PR #753, opened 2026-01-16
Call `modify-client-exec-fn-opts` in `eval.file`

pyrmont:
The `modify-client-exec-fn-opts` function is a mechanism for Conjure clients to manipulate the options that are passed to client functions. At present, the `eval.file` function does not use it. This commit changes that to make it more consistent with functions like `eval.eval-str`.

I want this so that I can silence log output for when a file is loaded.

### DIFF
--- a/fnl/conjure/eval.fnl
+++ b/fnl/conjure/eval.fnl
@@ -92,17 +92,6 @@
           (when f
             (f result)))))))
 
-(fn M.file []
-  (event.emit :eval :file)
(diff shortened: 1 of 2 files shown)


Olical:
Seems to work fine, I wonder _why_ it was like this, like did I just forget or was there a subtle reason lost to time. It seems to work great but I'll keep an eye out for any issues post merge. Thanks!

PR #757, opened 2026-02-01
Fix for issue #756.

russtoku:
This PR should address Issue #756.

The initial error reported led to discovery of a problem with the completion function for the Scheme, Guile, and Common Lisp clients. As it turns out, the `attempt to index local 'cursor_node' (a nil value)` message was the key to finding a solution.

The main problem lies with not calling `vim.treesitter.get_node` before the buffer is parsed by the parser method on `vim.treesitter.get_parser` on the current buffer for the filetype of the buffer. Luckily, there is already code in place to help so it takes just one line to resolve this issue. Also, this change will fix three Conjure client because they all call `get-completions-at-cursor` in `conjure.tree-sitter-completions`.

L.3 and L.86 are the main fix. The other changes are cosmetic whitespace removal.

### DIFF
--- a/fnl/conjure/tree-sitter-completions.fnl
+++ b/fnl/conjure/tree-sitter-completions.fnl
@@ -1,6 +1,5 @@
 (local {: autoload : define} (require :conjure.nfnl.module))
 (local a (autoload :conjure.nfnl.core))
-(local log (autoload :conjure.log))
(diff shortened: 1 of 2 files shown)


Olical (approved):
Makes sense! Tricky that we need to be aware of this every time we interact with tree sitter.

PR #766, opened 2026-03-16
JS/TS: Improved DX for transformation errors, proper multiline string handling, type import fixes, and other minor improvements.

Stansom:
Hello! I’ve fixed several crucial bugs affecting multiline string support. Previously, strings like `"Monday.\n Tuesday\n"` were flattened into a single line (e.g., `"Monday. Tuesday."`). Now, newlines are preserved as expected and are no longer transformed.

Regarding the TypeScript client, I fixed a bug where type imports — such as `import type { Result } from "error"` — were not functioning correctly. These imports now work as intended.

Finally, I addressed a major Developer Experience issue. Previously, when a transformation error occurred in an arrow function containing a forbidden keyword (like `this`), the code would throw an exception in the REPL and halt further transformation. This could lead to unexpected behavior. Now, the system instead issues a warning explaining that the function cannot be transformed due to potential `this` context loss, allowing execution to continue safely.

### DIFF
--- a/fnl/conjure/client/javascript/transformers.fnl
+++ b/fnl/conjure/client/javascript/transformers.fnl
@@ -133,7 +133,8 @@
            _ (node-handlers.default child code)))))
 
 (each [_ t (pairs
-             [:expression_statement
+             [:type_alias_declaration
+              :expression_statement
               :variable_declaration
               :return_statement
               :throw_statement
--- a/fnl/conjure/client/javascript/transformers.fnl
+++ b/fnl/conjure/client/javascript/transformers.fnl
@@ -95,15 +95,13 @@
   (let [body-node (tsc.get-child arrow-fn "body")
         forbidden (body-contains-forbidden-keyword? body-node code)]
     (if forbidden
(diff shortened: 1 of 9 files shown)


Olical (commented):
Looking good, great changes overall, thank you. I just think that node traversal can be simplified because I'm preeeety sure we already have the right node in memory in another function?

fnl/conjure/client/javascript/transformers.fnl
```diff
@@ -143,14 +163,14 @@
...
+              :function_declaration])]
 
   (set (. node-handlers t) handle-statement))
 
 (fn M.transform [s]
   (let [tree (tsc.get-tree s)
         root (tree:root)
         transformed (transform root s)]
-    (string.gsub transformed "%s*\n%s*" " ")))
+    (vim.fn.trim transformed)))
```
Olical, line 174: Oh nice!

Olical (approved):

Olical:
🎉 I'm gonna make a couple of small style / fennel changes but barely anything, LGTM!

Issue #767, opened 2026-03-20
Log Cursor Scroll Position Ignores Setting


delayedchoice:
First, thanks for this awesome development experience.  I default to clojure because of it.

Minor issue, my log scroll position stays on top even though it is set to bottom.
Here is my setup:
{
    "Olical/conjure",
    ft = { "clojure", "fennel" },
    lazy = true,
    init = function()
      vim.g["conjure#log#jump_to_latest#enabled"] = true
      vim.g["conjure#log#jump_to_latest#cursor_scroll_position"] = "bottom"
      vim.g["conjure#log#hud#enabled"] = false
    end,
  }

Thanks and keep up the good work!

delayedchoice:
Hey, thanks for your response!

To answer your question, I was looking for a behavior more like a terminal would behave, just continuously scrolling with new lines and if I want to see something earlier in the stream I scroll up.

I'm no vim power user. I wonder would it be possible to issue a zb like command from the code buffer without having to switch to the log buffer?

PR #768, opened 2026-03-27
Add last option to conjure#log#jump_to_latest#cursor_scroll_position

russtoku:
I've added a `last` option to the `conjure#jump_to_latest#cursor_scroll_position` configuration variable. This will place the cursor in the log buffer on the last line of the buffer. I've also (hopefully) clarified the description of the different options for the configuration variable.

This may address issue #767 but would need confirmation.


Olical:
Not required for that issue but hey, can't hurt 🙂 thanks Russ!

PR #773, opened 2026-05-08
Completion parse guards

seamusriordan:
This PR addresses two potential issues in completions which could lead to nvim errors.

1.  A continuation from [PR 772](https://github.com/Olical/conjure/pull/772).  If `parse!` fails, conjure will not attempt to query for completions and return an empty list instead.  Triggering the error is described in the above PR when multiple buffers are opened but not all have tree sitter languages available.

2. If a treesitter language is not installed conjure will not attempt to parse the completion query, a precursor to searching for completions.  I am able to trigger this by running `:TSUninstall <language>` and opening a file that will attempt tree sitter completions before the language is reinstalled.


Olical:
Looks good, thanks! These are nice improvements, I'll add some tests alongside your change too (I'm happy to do it since it's fiddly).

seamusriordan:
Thanks for getting both of these PRs merged!

PR #775, opened 2026-05-20
add picolisp stdio client

msladecek:
Hi, I recently implemented a picolisp plugin for conjure for my own use and I'd like to contribute it upstream.

Picolisp doesn't have any special syntactic features compared to other lisps, so implementing the plugin was fairly straight forward.

I started by copypasting the `conjure.client.fennel.stdio` plugin and adjusting the comment prefix, file suffix and repl command.

The only hiccup was that by default the picolisp repl responses contain [bracket pasting](https://en.wikipedia.org/wiki/Bracketed-paste) escape sequences and an extra `->` prefix ahead of the result, and I had trouble adjusting the plugin to eliminate them without also messing with `conjure.remote.client.stdio-rt`.

Finally, I figured it would be simpler to just run a simple custom read-eval-print-loop instead of trying to clean up the output of the built-in one.
So in the end, the plugin is almost totally identical to the  fennel stdio one.

I also added documentation files, also following  fennel-stdio as an example.

Please let me know if there is anything else to add/fix/cleanup, I would be happy to update the pull request as needed. 

Picolisp itself is available at: https://picolisp.com/wiki/?home


fnl/conjure/client/picolisp/stdio.fnl
```diff
@@ -0,0 +1,144 @@
...
+(when (config.get-in [:mapping :enable_defaults])
+  (config.merge
+    {:client
+     {:picolisp
+      {:stdio
+       {:mapping {:start "cs"
+                  :stop "cS"}}}}}))
+
+(local cfg (config.get-in-fn [:client :picolisp :stdio]))
+(set M.state (client.new-state #(do {:repl nil})))
```
Olical: Should `or` this with the existing state, that way if you reload this file the state is preserved and not wiped between file reloads in Conjure.

```fennel
(set M.state (or M.state (client.new-state #(do {:repl nil})))
```

fnl/conjure/client/picolisp/stdio.fnl
```diff
@@ -0,0 +1,144 @@
+(local {: autoload : define} (require :conjure.nfnl.module))
+(local core (autoload :conjure.nfnl.core))
+(local nfs (autoload :conjure.nfnl.fs))
```
Olical: Unused, we can drop this one.

fnl/conjure/client/picolisp/stdio.fnl
```diff
@@ -0,0 +1,144 @@
...
+
+(local cfg (config.get-in-fn [:client :picolisp :stdio]))
+(set M.state (client.new-state #(do {:repl nil})))
+
+(set M.buf-suffix ".l")
+(set M.comment-prefix "# ")
+(set M.form-node? ts.node-surrounded-by-form-pair-chars?)
+
+(fn M.comment-node? [node]
+  (node-prefixed-by-chars? node "#"))
```
Olical: Should be the following:

```fennel
(ts.node-prefixed-by-chars? node ["#"])
```

It's from the `ts` module and it takes a table of strings, not a single string.

fnl/conjure/client/picolisp/stdio.fnl
```diff
@@ -0,0 +1,144 @@
...
+  (with-repl-or-warn
+    (fn [repl]
+      (repl.send
+        (.. opts.code "\n")
+        (fn [msgs]
+          (when (and (= 1 (core.count msgs))
+                     (= "" (core.get-in msgs [1 :out])))
+            (core.assoc-in msgs [1 :out] (.. M.comment-prefix "Empty result.")))
+
+          (let [msgs (core.filter #(not= ".." (. $1 :out)) msgs)]
```
Olical: I _think_ the `".."` output is a Fennel specific thing and we don't need to filter it out here?

Olical (changes requested):
Looks good, just a couple minor things. A great start though! And thanks for updating the doc files etc too!

fnl/conjure/client/picolisp/stdio.fnl
msladecek: yep, two dots are printed by fennel's repl when an incomplete form is submitted so the user should continue submitting input. The  picolisp's repl doesn't do that. Removed the filter.

fnl/conjure/client/picolisp/stdio.fnl
msladecek: fixed

fnl/conjure/client/picolisp/stdio.fnl
msladecek: true, removed

fnl/conjure/client/picolisp/stdio.fnl
msladecek: makes sense, fixed

msladecek:
Thank you for the review @Olical, I applied the changes as you suggested.

Olical (approved):
LGTM!

PR #778, opened 2026-06-03
Add python virtual env tip, fix __name__ doc

frenchy64:
Coming from Clojure it wasn't obvious to me how to connect Conjure to a Python REPL within an actual project managed by e.g., Poetry. This doc should help people like myself figure it out more quickly.

I noticed the `__name__` snippet was incorrect so I fix that too.


Olical (approved):
Some good context, thanks! I think I overlook this part too easily. It's been MANY years since I called myself a "Python developer" but I guess the instincts of virtual environments being inherited by sub-processes was kind of natural and implied for me.

I also only use https://docs.astral.sh/uv/ these days for all of my Python so I'd end up configuring Conjure to use the `uv` command to start my REPL and then I never need to worry about virtual environments or my shell state either which is nice.

PR #781, opened 2026-07-13
Respect config in scheme when treesitter is disabled

seamusriordan:
Currently the scheme and guile clients will call tree-sitter facilities when validating a string to be evaluated by the repl even if `#extract#tree_sitter#enabled` is false in the config.  Furthermore the treesitter completions will also call some facilities if it is disabled (this has been harmless due to failure checks) but should also respect the config.

Added a `ts.enabled?` guard around `get-completions-at-cursor` with tests

Added a `ts.enabled?` check around `valid-str?` which returns true of tree-sitter is disabled with tests


russtoku:
After a quick look, it seems to me that this should work fine.

PR #784, opened 2026-08-23
Get def_word working with Fennel functions declared via λ/lambda

jship:
Fennel functions declared with `λ`/`lambda` have TS nodes of `lambda_form` rather than `fn_form`.


russtoku:
A quick peek shows that this will be a welcome addition. Thanks!

Olical:
Agreed!

PR #785, opened 2026-08-23
Improve hot reload for conjure.tree-sitter module

jship:
Use `define` for more consistent hot reloading


russtoku:
A quick look suggests that this should be good-to-go. Thanks!

PR #786, opened 2026-08-30
Add doc-str support to SQL client

jship:
Adds doc-str support for the SQL client. It digs around in treesitter to know which help command to send based on what's under the cursor. The default help command prefixes (`\d` for tables, `df+` for functions, and `\h` for statements) are tailored for `psql`. Hopefully the attached demo video is legible! I needed to compress it heavily to fit under GitHub's 10MB limit:

https://github.com/user-attachments/assets/2634c582-ae4c-47cb-97d3-0b70b9afce6c


russtoku:
Nice addition to make a SQL developer's life better!

I was a bit confused when watching your video so I'll need to actually use this new feature to clear up my confusion.

Olical:
Apologies for the delay, I was away running and attending (at the same time) my first stag party. I'll get this over the week as I catch up on work and sleep 😅

